PR #8, opened 2015-01-17
:lipstick: bunch of semicolons

muratsu:


bterlson:
Thank you for cleaning up my garbage!

Issue #9, opened 2015-01-22
Not escaping characters correctly


arv:
My input html has

```html
<p>1 &lt; 2</p>
```

and the output ends up being

```html
<p>1 < 2</p>
```

bterlson:
Do you have the latest version installed? Here's what I'm getting:

```html
<!doctype html>
<emu-clause>
  <h1>test</h1>
  <p>Test that a &lt; b escapes properly</p>
  <emu-alg>
    1. a &lt; b
  </emu-alg>
</emu-clause>
```
compiles to

```html
<!doctype html>
<head></head><body><div><h2>Table of Contents</h2><ol class="toc"><li><a href="#"><span class="secnum">1</span> test</a></li></ol></div><emu-clause>
  <h1><span class="secnum">1</span>test</h1>
  <p>Test that a
  &lt; b escapes properly</p>
  <emu-alg><ol>
  <li>a &lt; b</li>
</ol></emu-alg>
</emu-clause>
</body>
```
Which has the proper escaping. In Chrome dev tools it appears the characters are unescaped - perhaps that's the confusion?

arv:
My `&lt;` is not inside an `<emu-alg>`, but inside an `<emu-clause>`

https://github.com/arv/ecmascript-object-observe/blob/master/spec/KeyAlgorithmsAndSemantics.html#L45

http://arv.github.io/ecmascript-object-observe/

```
$0.outerHTML
"<p>An observation is conceptually a tuple:
  <observed object,="" observer="" function,="" accepted="" change="" types="">. Object.observe ensures that this tuple is represented as an observerRecord in the associated Notifier's ChangeObservers list.</observed></p>"
```

bterlson:
I had that above as well, but it turns out the key is if you put spaces before and after your angle bracket it will be escaped...

I guess this is because when processing paragraphs, I loop through each text node and pass its textContent to emd. At this stage I've lost the escaping of element-like things. I then set the innerHTML of a temporary element to this result and loop over the children of this temporary element, appending each in the right place.

I think I can fix this by first cloning the text node and appending it to a temporary div. This will let me call innerHTML on that temporary div, preserving the escapes. Anyone know a better way to do this that doesn't involve a temporary div? Is there any API in the DOM like textContent that returns escapes verbatim?

domenic:
Is there a reason why you can't just pass `pElement.innerHTML` to ecmarkdown?

domenic:
[...]

PR #13, opened 2015-03-17
Update to jsdom 4.x

### LINKED ISSUE #12: Update jsdom to remove native module dependency
Would be awesome

domenic:
Fixes #12


domenic:
Tacked on an Ecmarkdown update while I was at it. Might fix #9.

bterlson:
The only problem I have with this is dropping support for Node - I'd prefer to support all node-like hosts unless there's a good reason not to. Is there a way around it other than dropping jsdom?

domenic:
Not really. If it helps, just think of it as only supporting modern versions of Node, instead of year-old ones. (Node is to io.js as IE is to Spartan.)

michaelficarra:
@domenic That's disingenuous. Node 0.12 was released last month.

domenic:
Bump? Should I publish a fork?

bterlson:
Sad to lose node compat but I guess this is fine until the grand unification happens (hopefully) :)

domenic:
Awesome, thanks! New npm version please? :)

bterlson:
1.1.0 just for you, sir!

PR #48, opened 2015-09-22
more natural rotation of toggles in TOC

michaelficarra:
Now the arrow rotates all the way downward and retains its offset.


bterlson:
Whoops sorry, I missed this.

I suppose this is ok :) I kind of like the current style (aligns with Word, after all) but I like this too.

PR #50, opened 2015-09-24
Update emu-t and emu-nt for line breaks.

mysticatea:
From tc39/ecma262#64

I'm sorry, the `::after` way did not work as expected in IE.
So I had been investicating other approach.

I chose inserting a `<wbr>` and a `\u200B` (a zero-width space).

- IE11 makes new line at a `\u200B`, but not at `<wbr>`.
- Chrome makes new line at `<wbr>`, but not at a `\u200B` (2000-200B are the same. [it's a bug](https://code.google.com/p/chromium/issues/detail?id=60452)).
- Firefox makes new line at a `\u200B` or `<wbr>`.

It inserts those to after `<emu-t>` and `<emu-nt>`.
Because IE does not make new line if it inserted those to inside of `<emu-t>` and `<emu-nt>`. (I think it's a correct behavior)

### DIFF
--- a/css/elements.css
+++ b/css/elements.css
@@ -164,6 +164,7 @@ emu-rhs emu-nt {
 emu-t {
     font-family: monospace;
     font-weight: bold;
+    white-space: nowrap;
 }
 
 emu-production emu-t {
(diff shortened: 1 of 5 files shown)


mysticatea:
Please forget this PR.
I found more simple approach while I was investigating IE disabled `::after`.

PR #51, opened 2015-09-24
Update emu-t and emu-nt for line breaks

mysticatea:
From tc39/ecma262#64

The `::after` way did not work as expected in IE, because `<emu-t>` and `<emu-nt>` are `inline` element.
I modified these elements to `inline-block`, then browsers make new line at the border of blocks if it's overflowed.
This way needed resetting `text-indent`, because this style is inherited from parent and `<emu-rhs>` has `text-indent: -25px;`.

### DIFF
--- a/css/elements.css
+++ b/css/elements.css
@@ -149,8 +149,10 @@ emu-oneof {
 }
 
 emu-nt {
+    display: inline-block;
     font-style: italic;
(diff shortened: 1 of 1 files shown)


bterlson:
This looks good. Thank you so much @mysticatea!

PR #62, opened 2015-10-16
Fix all-inline rendering of productions in Note in 11.9.1

jmdyck:
... (Rules of Automatic Semicolon Insertion).

(see http://tc39.github.io/ecma262/#sec-rules-of-automatic-semicolon-insertion)


bterlson:
Nice!

PR #72, opened 2015-12-12
Link directly to <dfn>s if they have an ID

### LINKED ISSUE #68: When <dfn>s have IDs, autolinks to them should go to that ID, not to the parent clause
Some clauses are long and it would be good to go to the term being defined instead of having to Ctrl+F for it.

domenic:
Fixes #68.

This was trickier than expected because of the way autolinking always indirects through creating an emu-xref to the section whose innerHTML is `<a href="#section-id">term</a>`. So I had to extend emu-xref to also allow emu-xrefs directly to terms, with innerHTML `<a href="#term-id">term</a>`.

That seems kind of weird, and an alternate approach would be for autolinking to not create emu-xref wrappers for dfns, or at least not for dfns that have their own IDs. Let me know which you prefer.

### DIFF
--- a/lib/Dfn.js
+++ b/lib/Dfn.js
@@ -5,12 +5,21 @@ const parent = require('./utils').parent;
 
 module.exports = class Dfn extends Builder {
   build() {
-    const parentClause = parent(this.node, ['EMU-CLAUSE', 'EMU-INTRO', 'EMU-ANNEX']);
-    if (!parentClause) return;
+    const id = getClosestId(this.node);
+    if (!id) return;
 
     this.spec.biblio.terms[this.node.textContent] = {
-      id: parentClause.id,
+      id: id,
       location: ''
     };
   }
 };
+
+function getClosestId(node) {
+  if (node.id) return node.id;
+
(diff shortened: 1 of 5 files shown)


domenic:
I have a commit ready to upload that almost works and removes all emu-xref indirection from autolinks in favor of just `<a>`. Let me know if that sounds interesting.

(Right now for some reason it double-links forward references, but I can presumably fix that...)

bterlson:
I was getting the feeling that emu-xref would actually be the holder of any xrefing UI affordances so I was ok with the `<emu-xref><a>foo</a></emu-xref>` incantation. In general it seems ok to distinguish links from xrefs?

That said this change looks good, and I think I can take it as is can't I?

domenic:
This change is OK to take as-is if you are OK with extending the semantics of `<emu-xref>` to also go to dfns. The behavior implemented here is that `title` and `aoid` are ignored and the innerHTML is always used as the xref contents (with no fallback). If this is OK, we should probably document it in https://bterlson.github.io/ecmarkup/#emu-xref.

PR #71, opened 2015-12-12
Code cleanups

domenic:
No behavior changes

PR #73, opened 2015-12-12
WIP dfn.js-style reference links

domenic:
This is running into styling troubles trying to get the popup to correctly size itself and appear in the right place and toggle on/off. Since I hear you are making changes to related style areas anyway, i.e. the .utils class, I thought instead I'd just upload what I have. Most of the build-side code is ready for review, although the client-side code is not.

On top of #71 since I use some new jsdom features (classList). Best to review the last commit only.

### DIFF
--- a/bin/ecmarkup.js
+++ b/bin/ecmarkup.js
@@ -12,7 +12,7 @@ const readFile = Promise.promisify(fs.readFile);
 const utils = require('../lib/utils');
 const debounce = require('promise-debounce');
 
-const jsDependencies = ['menu.js', 'findLocalReferences.js'];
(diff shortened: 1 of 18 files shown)


domenic:
Also, this only works on things with `.utils` descendants right now, i.e. emu-clauses. I think we should make it work for anything with an ID. There are a few approaches to this, but I think the best is to add a `span.utils` in an appropriate place for everything that has an ID. `dfn`s, `emu-table`s inside their `caption`, and anything else we can think of.

bterlson:
Find references is implemented in 3.10.

PR #78, opened 2016-01-14
skip over <ins> and <del> parents in utils.shouldInline

michaelficarra:
Took me a little while to figure out why my production was being inlined.


bterlson:
Good catch! You should ping me on IRC with issues when you find them, I can probably save you many hours of hair-pulling :)

PR #82, opened 2016-03-16
fixes #81: support ins/del diff styles for table rows

### LINKED ISSUE #81: support ins/del diff styles for table rows
So I don't have to put the entire table in a &lt;del> and the entire table with one row changed in an &lt;ins>. PR incoming.

michaelficarra:
Fixes #81. :rowboat: :rainbow:

### DIFF
--- a/css/elements.css
+++ b/css/elements.css
@@ -373,6 +373,10 @@ del.block {
     display: block;
 }
 
+tr.ins > td > ins, li.ins > ins {
+  border: 0;
+}
+
 /* Menu Styles */
 #menu-toggle {
   font-size: 2em;
--- a/css/elements.css
+++ b/css/elements.css
@@ -354,7 +354,7 @@ table.lightweight-table td, table.lightweight-table th {
 }
 
 /* diff styles */
-ins, tr.ins > td {
+ins, tr.ins > td, li.ins {
     background-color: #e0f8e0;
     text-decoration: none;
     border-bottom: 1px solid #396;
@@ -364,7 +364,7 @@ ins.block {
     display: block;
 }
 
-del, tr.del > td {
+del, tr.del > td, li.del {
     background-color: #fee;
     text-decoration: line-through;
 }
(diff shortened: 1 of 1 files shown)


michaelficarra:
I added support for lists as well. Some related information from the HTML specification (basically saying ins/del are incompatible with tables and lists):

* [4.6.5 Edits and lists](https://www.w3.org/TR/html5/edits.html#edits-and-lists)
* [4.6.6 Edits and tables](https://www.w3.org/TR/html5/edits.html#edits-and-tables)

domenic:
-1. You should not mark these up with unsemantic .del and .ins classes. You should use `<ins>` and `<del>` on the contents of the table. See e.g. https://domenic.github.io/zones/#monkeypatch-realms

michaelficarra:
@domenic Sorry, but HTML's not expressive enough for us here (without potentially some huge hoop-jumping). You simply can't reproduce the same aesthetic with `<ins>` in every `<td>`. I would be okay with the compromise of using the `ins` class on the `<tr>` for styling, putting an `<ins>` in every `<td>`, and disabling styles for `tr.ins > td > ins`.

domenic:
I don't understand why your usage can't just use my technique.

michaelficarra:
@domenic I'm trying to encode the information that the entire row was added, as well as allow ecmarkup to style that row in a way that makes it clear to humans. See the difference in appearance between my linked example and yours. I can't do that without a class on the `<tr>` because of HTML restrictions in the structure of `<table>`s (and `<ul>` and `<ol>`).

domenic:
I think it's pretty clear that in my example an entire row was added.

michaelficarra:
@domenic I've added the style reset to `<ins>` within tables/lists, so use of this class is entirely optional. Using individual `<ins>` to mark up semantic insertion will still be encouraged, though proposal authors will *also* be able to mark a row or list item as being inserted. I think this patch now provides strictly additive value.

bterlson:
Sorry I didn't see a notification for this PR anywhere :( Am I even watching my own repo??

Anyway, I think this is fine. Long term I hope diff tools like [aria's](i-a.github.io/ecma262-compare/) will make manual ins/del less used. But for now it seems ok to add some authoring conveniences even though it's not as ideal as it could be.

PR #89, opened 2016-05-05
Printing styles and oldtoc styles to conform with ECMA PDF files requirements

caridy:
Most of the tricks are from the following doc:
https://www.smashingmagazine.com/2015/01/designing-for-print-with-css/

### DIFF
--- a/css/elements.css
+++ b/css/elements.css
@@ -548,3 +548,41 @@ tr.ins > td > ins, li.ins > ins {
     margin: 0; padding: 0;
   }
 }
+
+/* extra styles for oldtoc */
+body.oldtoc .anchor {
+  display: none;
+}
+
+body.oldtoc emu-annex h1 {
+  text-align: center;
+  font-size: 1.6em;
+}
+
+body.oldtoc emu-annex h1 .secnum {
+  display: block;
+  left: 0;
+  text-align: center;
+  position: relative;
+}
+
+body.oldtoc emu-annex h1 .secnum::before {
+  content: "Annex ";
+}
+
+body.oldtoc emu-annex h1 .secnum::after {
+  display: block;
+}
+
+/* extra styles for printer */
+emu-annex h1,
+emu-intro h1,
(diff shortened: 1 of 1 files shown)


css/elements.css
```diff
@@ -548,3 +548,41 @@ tr.ins > td > ins, li.ins > ins {
     margin: 0; padding: 0;
   }
 }
+
+/* extra styles for oldtoc */
+body.oldtoc .anchor {
+  display: none;
```
caridy: hiding the `#` symbol from the UI for PDF files

css/elements.css
```diff
@@ -548,3 +548,41 @@ tr.ins > td > ins, li.ins > ins {
...
+}
+
+body.oldtoc emu-annex h1 {
```
caridy: the following 3 rules are just to add the word Annex as a prefix, and centering the secnum. We could probably do this when generating the secnum though.

css/elements.css
```diff
@@ -548,3 +548,41 @@ tr.ins > td > ins, li.ins > ins {
...
+
+body.oldtoc emu-annex h1 .secnum::after {
+  display: block;
+}
+
+/* extra styles for printer */
+emu-annex h1,
+emu-intro h1,
+emu-toc div {
+  page-break-before: always;
```
caridy: this will guarantee to break before any of those element. unfortunately the trick does not work for custom elements, that's why I have to use h1 or any other first element, we could probably do some more elegant solution with :first-child or something.

css/elements.css
```diff
@@ -548,3 +548,41 @@ tr.ins > td > ins, li.ins > ins {
...
+}
+
+h1, h2, h3, h4, h5,
+emu-figure > figure,
+emu-table > figure {
+  page-break-inside: avoid;
```
caridy: same as above, but this one is just preventing those elements to be the last element of a page, breaking to the next when in that situation.

PR #99, opened 2016-08-24
Migrate ecmarkup to TypeScript

rbuckton:
Migrates ecmarkup to TypeScript. I've set up the tsconfig.json to use some fairly strict settings (noImplicitAny, strictNullChecks, etc.), though they can be relaxed if necessary.


.vscode/tasks.json
```diff
@@ -0,0 +1,10 @@
+{
+    // See https://go.microsoft.com/fwlink/?LinkId=733558
```
bterlson: Does vs actually allow comments in JSON? Evil!

rbuckton: Yep. Only for JSON config files, so that various options can be documented. The comments are stripped before the JSON is parsed by the editor.

bterlson: Boo. Not JSON then. Should be using a different extension.

bterlson:
Let's put @internal on its own line (when used before class methods, at least)

domenic:
Why change from 2-space indent to 4-space indent? It makes the diffs useless and breaks blame.

rbuckton:
The TypeScript compiler emits 4 space indentation. There's no setting to override it at present.

rbuckton:
I'm considering adding `lib` to .gitignore, since it now only contains build outputs.

domenic:
I see, yes, build output should not be included in the repo... it will just confuse people like it confused me.

bterlson:
Agreed, let's not check in the built output.

bterlson:
Looks good with the latest commits! Thanks @rbuckton!

PR #101, opened 2016-08-30
Fix CI build

mathiasbynens:


bterlson:
was just setting out to do this myself... thanks :-D

PR #103, opened 2016-09-01
Patch __awaiter to use bluebird

rbuckton:
Keep in mind that even with this patch, the return **type** of an async function will still be the ES6 Promise, even though the runtime value will be a bluebird Promise.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
   "author": "Brian Terlson",
   "license": "MIT",
   "dependencies": {
+    "bluebird": "^3.4.5",
     "chalk": "^1.1.1",
     "ecmarkdown": "^3.0.9",
     "grammarkdown": "^1.0.5",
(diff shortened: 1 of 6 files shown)


domenic:
Why? This seems like crazy black magic. It would be nicer to get rid of Bluebird and just use native promises?

rbuckton:
@bterlson requested it. I'd prefer to use native promises as well.

domenic:
I guess it would be good to at least add a comment linking to the docs on `__awaiter`.

rbuckton:
@domenic Added in de98f8c

bterlson:
@domenic the key reason is perf - using bluebird and the __awaiter override shaves ~3s off the ECMA262 build process. Will document it better.

bterlson:
Committed some extra rationale in 33da4fb. Closing this as I've incorporated all the changes here as part of my previous commits.

bterlson:
Make that 662f3a2.

PR #113, opened 2017-01-21
override padding-left and margin-left in inline emu-production

michaelficarra:
Before:

screen shot 2017-01-21 at 11 00 22

After:

screen shot 2017-01-21 at 10 59 57

### DIFF
--- a/css/elements.css
+++ b/css/elements.css
@@ -167,19 +167,16 @@ emu-production {
     margin-left: 5ex;
 }
 
-
 emu-grammar.inline, emu-production.inline,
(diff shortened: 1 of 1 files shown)


michaelficarra:
Also, the file is inconsistently indented, so I just went with 4-space. It should probably be reformatted.

bterlson:
Nice find!

Also, yes, that css file is a dumpster fire. Like, majorly. And I'm really sorry about it.

bterlson:
I'll do an npm deploy later today!

PR #134, opened 2018-03-21
Hide #menu-toggle for print styles

mathiasbynens:
This makes the print output look good. Print PDFs can now be generated with the following command:

```sh
chrome --headless --disable-gpu --print-to-pdf 'https://tc39.github.io/ecma262/'
```

### DIFF
--- a/css/elements.css
+++ b/css/elements.css
@@ -898,3 +898,9 @@ li.menu-search-result-term:before {
   text-align: right;
   padding-right: 5px;
 }
+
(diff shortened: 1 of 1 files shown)


mathiasbynens:
To get this change live on https://tc39.github.io/ecma262/, two things are needed: a new ecmarkup release must be published, and a commit must be pushed to the tc39/ecma262 so the generated spec gets updated.

PR #137, opened 2018-04-02
Allow custom copyright boilerplates. (Enables use for non-Ecma specs)

rwaldron:
I want to write Johnny-Five related spec documents and I want to use ecmarkup, but right now I can't. This feature enables using ecmarkup for non-Ecma specifications; and while that may sound a little self-contradictory, ecmarkup is a great tool for writing and generating specification documents, and sometimes those specifications don't have a standards body (yet?) or will never become an Ecma spec (for whatever reason). Anyway, this is useful.

### DIFF
--- a/src/Spec.ts
+++ b/src/Spec.ts
@@ -448,7 +448,7 @@ export default class Spec {
         return;
       }
 
-      node = walker.nextNode();
+      node = walker.nextNode()!;
     }
   }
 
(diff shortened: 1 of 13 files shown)


ljharb:
ooh, does this mean that the ecma262 boilerplate could actually live inside ecma262 instead of in ecmarkup?

rwaldron:
@ljharb yep!

bterlson:
This is very nice, thank you Rick!

Issue #138, opened 2018-04-04
Detect and signal when duplicate ids occur


rwaldron:
Ref: https://github.com/tc39/ecma262/pull/1141#issuecomment-378646964

PR #139, opened 2018-04-06
Detect and warn when duplicate ids exist and --verbose flag is present. Fixes gh-138

rwaldron:


bterlson:
Looks awesome, thanks @rwaldron!

PR #145, opened 2018-05-24
Switch to Grammar.convert and sync code paths

rbuckton:
- Upgrade grammarkdown to 2.0.11 (only interesting change was inclusion of `Grammar.convert` static method)
- Switch to calling `Grammar.convert` (performs synchronous parse and emit to a string)
- Switch `Grammar.enter`, `Algorithm.enter`, and `walk` back to sync execution.


rbuckton:
@bterlson: Can you take a look?

PR #148, opened 2018-10-12
Insert a selectable space between heading number and text

gibson042:
I'm sick of selecting text like "20.3.1.1Time Values and Time Range".


gibson042:
ping @bterlson

PR #167, opened 2020-01-10
Update es6biblio.json (2020-01-09).

rkirsling:
Generated at https://github.com/tc39/ecma262/commit/63be86408b2c01899d6fd565db9a85d76a56de42.


michaelficarra:
Maybe we should rename it to something more appropriate than `es6biblio.json` while we're at it?

rkirsling:
Good call!

ljharb:
@rkirsling now that tests are fixed on master and enabled to run on PRs, i can see that this PR breaks tests. can you update it to pass them?

rkirsling:
Fixed!
(Beyond `tc39.github.io` → `tc39.es`, changes include a few updated `id`s and `abstract operations` having become a 262 reference.)

PR #169, opened 2020-01-12
Add CLI option ecma-262-biblio.

### LINKED ISSUE #168: 262 needs a way to opt out of loading its own bibliography
We need a real solution to https://github.com/tc39/ecma262/pull/1839.

This may just amount to introducing a CLI flag that lets one skip calling `loadES6Biblio` in Spec.ts.

rkirsling:
Resolves #168. Depends on #167.

Add boolean flag to skip loading ecma262biblio.json (previously es6biblio.json).

Also delete bin/args.js which seems to be obsolete (became unused [here](https://github.com/bterlson/ecmarkup/commit/c966f5b9a444850e3bfc736cf50c9091ff4ef33d#diff-5ca093f6985c457c5e2f010561b91582) and then further incorrect [here](https://github.com/bterlson/ecmarkup/commit/c966f5b9a444850e3bfc736cf50c9091ff4ef33d#diff-5ca093f6985c457c5e2f010561b91582)).


src/args.ts
```diff
@@ -7,6 +7,7 @@ const args = nomnom.script('ecmarkup')
     help: { abbr: 'h', flag: true, help: 'Display this help message' },
     watch: { abbr: 'w', flag: true, help: 'Rebuild when files change' },
     biblio: { abbr: 'b', metavar: 'FILE', help: 'Write a biblio file to FILE' },
+    omitECMA262Biblio: { full: 'omit-ecma262biblio', flag: true, default: false, help: 'Skip loading ECMA-262 biblio file' },
```
michaelficarra: Couldn't it be

```js
ecma262biblio: { full: 'ecma-262-biblio', flag: true, default: true, help: 'Load ECMA-262 biblio file' },
```

and then from the CLI we do `--no-ecma-262-biblio`?

rkirsling: Ah, that does seem better indeed. Will do!

rkirsling: Done!

src/Spec.ts
```diff
@@ -147,6 +147,10 @@ export default class Spec {
       this.opts.copyright = true;
     }
 
+    if (!this.opts.hasOwnProperty('ecma262Biblio')) {
```
ljharb: wouldn't this break if someone passed a `hasOwnProperty` op?
```suggestion
    if (!Object.prototype.hasOwnProperty.call(this.opts, 'ecma262Biblio')) {
```

rkirsling: Hmm, I was just copying the style of the surrounding code though.

You are right that calling `ecmarkup` with `--hasOwnProperty foo` does give:
```
TypeError: this.opts.hasOwnProperty is not a function
    at new Spec (/Users/rkirsling/Documents/GitHub/ecmarkup/lib/Spec.js:75:24)
```
...but then it's a pre-existing issue.

I can file an issue to have ecmarkup properly guard against unexpected options across the board?

ljharb: no need for an issue; but sure, a separate PR to fix all such cases, with tests, seems great

rkirsling: Tried just switching these to typeof undefined checks since it's simple enough to tack on here and `nomnom` doesn't seem to give a simple way to ban unrecognized arguments; what do you think?

Not sure how to test this easily since existing tests are all oriented around build output...but this does ensure that a `hasOwnProperty` option is ignored instead of causing a TypeError.

PR #170, opened 2020-02-01
Do not auto-link [Await]

bakkot:
Fixes https://github.com/tc39/ecma262/issues/1836.

(Two commits so you can see that it does in fact fix the issue, though they should be squashed into one for landing.)


ljharb:
Should we also add a test that ensures that `Await()`s are still autolinked?

bakkot:
I guess we could? I don't think it's really necessary in light of the existing tests.

PR #171, opened 2020-02-01
Enforce, fix, and modernize linting

bakkot:
`npm run lint` existed, but was unused (and broken). This PR adds it to `npm run test` so it will get enforced by CI, fixes it, adds linting to typescript files, and bumps eslint.

I haven't changed any of the linting options (at least, not intentionally), though there's some options I should certainly like to add. But we can bikeshed those in another issue or PR.


bin/.eslintrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  extends: '../.eslintrc.js',
```
ljharb: this should be done with "overrides" in the root, rather than having more than one eslint config file.

.eslintrc
```diff
@@ -1,45 +0,0 @@
-{
```
ljharb: i would very much prefer to retain the `.eslintrc` format rather than using JS for config

bakkot: Moved to `.eslintrc.json`.

bin/.eslintrc.js
bakkot: Done.

PR #176, opened 2020-03-12
Add emu-not-ref

### LINKED ISSUE #158: suppress auto-linking
Sometimes we don't want ecmarkup to auto-link a word. For example, see https://github.com/tc39/ecma262/issues/1730 and https://github.com/tc39/ecma262/pull/1731.

It would be nice if ecmarkup would:
- suppress auto-linking within the `<div id=metadata-block>` element; and/or
- allow the document to spot-disable auto-linking via some markup (e.g.: `<span class="do-not-autolink">Mailing List Archives</span>`)

bakkot:
Fixes #158 by adding a new tag, `emu-not-ref`, which can be used to explicitly suppress auto-linking to definitions.

This tag still shows up in the HTML output, with no semantics. I feel a little bad about that but don't really want increase compile times by post-process'ing it out.

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -248,6 +248,31 @@ toc: false
   <p><emu-xref aoid="Get"></emu-xref> is an abstract operation from ES6</p>
 </emu-clause>
 
+<emu-clause id="emu-not-ref" namespace="emu-not-ref">
+  <h1>emu-not-ref</h1>
+  <p>Suppresses automatic linking to definitions.</p>
+
+  <h2>Example</h2>
+  <pre><code class="language-html">
+  &lt;div id="not-ref-section-1">
+    &lt;p>An &lt;dfn>example&lt;/dfn> is used for illustrative purposes.&lt;/p>
+  &lt;/div>
+  &lt;div id="not-ref-section-2">
+    &lt;p>When a word defined in another section (or algorithm step) is used, as in this example, it is normally automatically linked to the section containing the definition.&lt;/p>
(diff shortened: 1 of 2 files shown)


ljharb:
Are the build times that slow that it's an issue to remove it?

spec/index.html
```diff
@@ -248,6 +248,31 @@ <h2>Example</h2>
...
+    &lt;p>When such a word should not be automatically linked, for &lt;emu-not-ref>example&lt;/emu-not-ref> when using its colloquial definition, it can be wrapped with this tag to surpress the automatic linking.&lt;/p>
+  &lt;/div>
+  </code></pre>
+
+  <b>Result</b>
+  <div id="not-ref-section-1">
+    <p>An <dfn>example</dfn> is used for illustrative purposes.</p>
+  </div>
+  <div id="not-ref-section-2">
+    <p>When a word defined in another section (or algorithm step) is used, as in this example, it is normally automatically linked to the section containing the definition.</p>
```
michaelficarra: Shouldn't the word "example" be auto-linked here?

bakkot: It will be once this gets rendered! This document is written in ecmarkup.

bakkot:
> Are the build times that slow that it's an issue to remove it?

They're on the order of fifteen seconds on my computer, which is not slow. So, I'm going to go with yes.

Anyway, I guess we leave all the other `emu-` tags in. Maybe we should remove them but that can be a different PR.

spec/index.html
ljharb: :-p too bad the extension implies it's HTML

PR #178, opened 2020-03-17
collapsible pin pane

peterzhangsnail:
Add an icon to the pin pane to make it collapsable.
Here's how it may look on the ecma doc:

Closes https://github.com/tc39/ecma262/issues/1898


js/menu.js
```diff
@@ -848,3 +848,33 @@ document.addEventListener('DOMContentLoaded', function () {
   Toolbox.init();
   referencePane.init();
 })
+
+//add toggle icon to the PIN pane and add the click event handler to it
+window.onload = function(){
+  var MenuPins = document.getElementById('menu-pins');
+  var MenuPaneHeader = MenuPins.querySelectorAll('.menu-pane-header')[0];
```
ljharb:
```suggestion
  var MenuPaneHeader = MenuPins.querySelector('.menu-pane-header');
```

js/menu.js
```diff
@@ -848,3 +848,33 @@ document.addEventListener('DOMContentLoaded', function () {
   Toolbox.init();
   referencePane.init();
 })
+
+//add toggle icon to the PIN pane and add the click event handler to it
```
ljharb: Is there any way the PIN toggle could be added just like the current toggles are, so we don't have to duplicate the functionality?

js/menu.js
```diff
@@ -848,3 +848,33 @@ document.addEventListener('DOMContentLoaded', function () {
...
+
+  var MenuPinsList = document.getElementById('menu-pins-list'); //our click event handler toggles its display style between 'none' and 'block'
+
+  var MenuPaneHeaderToggle = document.createElement('span');
+  var MenuPaneHeaderToggleIcon = document.createTextNode('◢'); //we use this character as the toggle icon
+
+  MenuPaneHeaderToggle.appendChild(MenuPaneHeaderToggleIcon);
+  MenuPaneHeaderToggle.setAttribute('class','MenuPinsNormal MenuPinsNormalCollpase');
+  
+  MenuPaneHeaderToggle.onclick = MenuPinsToggle.bind(null,MenuPinsList,MenuPaneHeaderToggle); //add click event handler
```
ljharb: this should use addEventListener instead of onclick

js/menu.js
```diff
@@ -848,3 +848,33 @@ document.addEventListener('DOMContentLoaded', function () {
...
+
+  MenuPaneHeader.appendChild(MenuPaneHeaderToggle);
+};
+
+
+//click event handler
+function MenuPinsToggle(MenuPinsList,MenuPaneHeaderToggle){
```
ljharb:
```suggestion
function MenuPinsToggle(MenuPinsList,MenuPaneHeaderToggle) {
```

js/menu.js
```diff
@@ -848,3 +848,33 @@ document.addEventListener('DOMContentLoaded', function () {
...
+  if(MenuPinsList.style.display === 'none'){
```
ljharb:
```suggestion
  if (MenuPinsList.style.display === 'none') {
```

js/menu.js
```diff
@@ -848,3 +848,33 @@ document.addEventListener('DOMContentLoaded', function () {
...
+    MenuPinsList.style.display = 'block';
+    MenuPaneHeaderToggle.setAttribute('class','MenuPinsNormal MenuPinsNormalCollpase');
+  }else{
```
ljharb:
```suggestion
  } else {
```

js/menu.js
```diff
@@ -848,3 +848,33 @@ document.addEventListener('DOMContentLoaded', function () {
...
+
+  MenuPaneHeader.appendChild(MenuPaneHeaderToggle);
+};
+
+
+//click event handler
+function MenuPinsToggle(MenuPinsList,MenuPaneHeaderToggle){
+  if(MenuPinsList.style.display === 'none'){
+    MenuPinsList.style.display = 'block';
+    MenuPaneHeaderToggle.setAttribute('class','MenuPinsNormal MenuPinsNormalCollpase');
```
ljharb:
```suggestion
    MenuPaneHeaderToggle.setAttribute('class', 'MenuPinsNormal MenuPinsNormalCollapse');
```

js/menu.js
```diff
@@ -848,3 +848,33 @@ document.addEventListener('DOMContentLoaded', function () {
...
+  var MenuPins = document.getElementById('menu-pins');
+  var MenuPaneHeader = MenuPins.querySelectorAll('.menu-pane-header')[0];
+
+  var MenuPinsList = document.getElementById('menu-pins-list'); //our click event handler toggles its display style between 'none' and 'block'
+
+  var MenuPaneHeaderToggle = document.createElement('span');
+  var MenuPaneHeaderToggleIcon = document.createTextNode('◢'); //we use this character as the toggle icon
+
+  MenuPaneHeaderToggle.appendChild(MenuPaneHeaderToggleIcon);
+  MenuPaneHeaderToggle.setAttribute('class','MenuPinsNormal MenuPinsNormalCollpase');
```
ljharb:
```suggestion
  MenuPaneHeaderToggle.setAttribute('class', 'MenuPinsNormal MenuPinsNormalCollapse');
```

css/elements.css
```diff
@@ -912,3 +912,19 @@ li.menu-search-result-term:before {
...
+  cursor: pointer;
+  position: absolute;
+  right:5px;
+  display: inline-block;
+  width: 21.6px;
+  transform-origin: center;
+  transform: rotate(45deg);
+}
+
+.MenuPinsNormalCollpase {
```
ljharb:
```suggestion
.MenuPinsNormalCollapse {
```

js/menu.js
peterzhangsnail: That's good thinking. I'll open a new PR after figuring it out.

peterzhangsnail:
@ljharb Here's [the new PR](https://github.com/bterlson/ecmarkup/pull/183): use the method already existed to achieve it, just like your suggestion, looking for more tests and ideas.

js/menu.js
ljharb: Please reuse existing PRs instead of creating clutter by opening duplicate ones :-/

peterzhangsnail:
@ljharb  Done!

bakkot:
I feel like the orientation of the arrows in the screenshot is the opposite of what I expect. Specifically, I expect "down" to mean "expanded" in this context. Compare Github's `<details>`:

<details>example</details>

ljharb:
Ah, good call - i didn't notice before, but you're right. The triangles should match the way the TOC sections already work - which is as you describe.

@peterzhangsnail, can you fix?

peterzhangsnail:
@ljharb I run my build and the orientation of the button seems correct. Here's my process below:
(OS: win7 / build tool: git-bash)
build

peterzhangsnail:
@ljharb It's also easy to 'fix' anyway. Just a little exchange of strings in elements.css.

ljharb:
@peterzhangsnail when "closed", the arrow should be pointing left/inwards (at the title), not right/outwards.

rkirsling:
@ljharb It's true that an right-facing arrow on the right often means "go to detailed view" but I'm not sure I can think of an example of an left-facing arrow on the right being used for a closed drawer?

I think when it's on the right, the norm is to have the "closed" arrow point _upwards_—see the "pending reviewers" box on this very page, as well as the Preferences pages for Chrome and FF.

ljharb:
@rkirsling i'd expect "up/down", or "inwards/downwards". Chrome's preference arrows, at least, aren't expanders, they're navigation "go to"s.
<details>
<summary>this arrow points inwards/downwards, for example</summary>
</details>

rkirsling:
Yep, I agree with your rationale, I just don't recall seeing an inward-facing arrow on the right.

Chrome and FF Preferences have drawers, you just have to look around for them:
image

image

ljharb:
Ah, true. Perhaps the "pins" arrow should be on the left, then?

peterzhangsnail:
@ljharb That doesn't match with the toc header below it if we put the arrow on the left.

peterzhangsnail:
@ljharb Here's the upward/downward version of pin arrow: 
(Better change the transition time to 0s, a 180-degree rotate with non-zero transition time seems weird.)

ljharb:
@peterzhangsnail those are backwards too; the downwards arrow in an up/down is the one that shows the contents; upwards hides them.

peterzhangsnail:
@ljharb Click a downward/upward arrow then it shows/hides the content. You must click it first.

peterzhangsnail:
@ljharb Here's a gif to show the process:
GIF

ljharb:
Yes, that's backwards. When the content is hidden, the arrow should reflect that by pointing upwards. Clicking it should show the content and thus change it to point downwards, indicating that the content is visible (and that clicking it will make it invisible).

peterzhangsnail:
@ljharb Hope this is what we want: 
GIF

peterzhangsnail:
@ljharb Hope this is what we want:
GIF

peterzhangsnail:
@ljharb I have run my build routine with the new commit, everything works just fine.

ljharb:
The things we have to decide between:
1. disclosure triangle is on the left, closed === points inwards
  - Pros: consistency with TOC triangles
  - Cons: P in "Pins" does not align with "T" in "Table of Contents"
2. disclosure triangle is on the right, closed === points inwards
  - Pros: "closed" look is consistent with TOC triangles
  - Cons: right-alignment is inconsistent with TOC triangle alignment
3. disclosure triangle is on the right, closed === points upwards
  - Pros: ?
  - Cons: right-alignment is inconsistent with TOC triangle alignment; "closed" state is inconsistent with TOC triangles

My preference is 1, 2, 3.

rkirsling:
I don't have a strong opinion here, but I'll note that the upward/downward approach used by GitHub/Chrome/Firefox has downward as the closed state (_I was momentarily confused by this myself, hence all the edits to my first comment above_).

ljharb:
@rkirsling in <details></details> ?

rkirsling:
@ljharb What? That's not upward/downward.

ljharb:
@rkirsling right; so what are you referring to that's up/down in those browsers? `<details>` is the only thing i'm aware of in html.

rkirsling:
@ljharb I just re-referenced what I described in my first two comments: the screenshots as well as the "pending reviewers" box on this page.

ljharb:
@rkirsling ah, thanks, i see what you mean. In those cases they're not triangles, they're chevrons, and i think the difference, while subtle, actually matters quite a bit.

rkirsling:
Hmm, that's fair. Maybe that means it's confusing to have a filled triangle flip instead of rotate then? 🤔 In any event, I have no qualms with your order of preference.

PR #182, opened 2020-03-21
Less typescript syntax

bakkot:
Three commits:

- get rid of `import foo = require('foo')`, which is a typescript-specific abomination which is typically synonymous with `import * as foo from 'foo'`.
- when importing something just for its type, use `import type` (enforced with `"importsNotUsedAsValues": "error"` in `tsconfig.json`, described [here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html))
- sort `import type` imports first.

With these (in fact with just the first one) this compiles with the just-released babel v7.9.0 using

```json
{
  "presets": ["@babel/preset-typescript"],
  "plugins": ["@babel/plugin-transform-modules-commonjs"]
}
```

although this PR does not propose to switch compilation to babel.


src/Algorithm.ts
```diff
@@ -1,6 +1,7 @@
+import type { Context } from './Context';
+
 import Builder from './Builder';
-import emd = require('ecmarkdown');
-import { Context } from './Context';
+import * as emd from 'ecmarkdown';
```
ljharb: this isn’t actually the same; CJS has only a default export. TS’ module system is broken unless synthetic imports and esModuleInterop are enabled, at which point this would just be `import emd from ‘ecmarkdown’`

bakkot: These compile to literally identical code (below).

I don't know what it means to say "CJS only has a default export"; they are entirely different module systems.

```js
const emd = require("ecmarkdown");
const html = emd.document(contents);
```

ljharb: The thing you import * has to be a frozen module namespace object; this isn’t it. Thus, the import * syntax isn’t doing what the spec says it should, unless you enable the two config settings i mentioned.

bakkot: This program does not observe that difference, so who's to say that it's not frozen? The observable semantics are the same as if it were, and the spec only puts requirements on observable semantics.

ljharb: ecmarkdown, being CJS, exports an object - a single thing. A proper interop does not do what Babel does, and destructure that object into exports.

I agree that in this code, it seems fine, but no CJS module has named exports, so I’m not sure why this format is acceptable to you.

bakkot: > I agree that in this code, it seems fine, but no CJS module has named exports, so I’m not sure why this format is acceptable to you.

CJS modules don't have ES exports at all. They are different module systems entirely. Something has to hook them up somehow, and making the decision to hook them up this way is as valid as any other.

> (also, the spec dictates observable semantics, not observed semantics, so it shouldn’t matter what this code is doing if it’s possible for it to be observed to be incorrect)

Observable semantics is a whole-program property, not a property of individual lines of code. If the whole program has the correct observable behavior, then tautologically it not possible for it to be observed to be incorrect.

You can pretend that the compiler is first running an analysis to prove that this detail is not observed in this program, if you like. It isn't, to my knowledge, but you can think of that as being an analysis which happens to have a bug. As long as we're not relying on that bug - which we are not, since we are in the case where it is, in fact, not observed - I don't see what the problem is.

ljharb: By using TS's broken interop here, someone might, in the future, mutate the thing that's supposed to be a namespace object. Correctness isn't just about the current state of the program, it's about all future states.

In other words, TS's default choice here is *not* valid, and that's why `tsc init` has set both the esModuleInterop and synthetic whatever flags for a long time - because the TS team also knows that's the only compliant way forward. It's only legacy projects, like this one, that have to make the explicit decision to opt into the correct TS semantics.

bakkot: > By using TS's broken interop here, someone might, in the future, mutate the thing that's supposed to be a namespace object.

Not without explicitly disabling TypeScript's type enforcement which prohibits you from doing so.

In any case, we can just reject PRs which rely on that particular compiler bug. I _really_ don't see what the issue is here.

---

Similarly, by using JavaScript instead of Haskell in this project someone might, in the future, change a value which was intended to be immutable. Should we therefore switch to Haskell? Usually it is recognized that a program having the intended semantics without reliance on compiler bugs is enough, even though it is possible that someone might later change the program to be incorrect.

Anyway, if you want to send a PR which turns on that flag, I would be happy to review it. I don't know if that would entail a major version bump, but even if it does I don't think that would be a problem. In the mean time, this is a strict improvement over the current state of affairs. I would like to merge it as-is unless there's something else wrong with it.

ljharb: It wouldn't be an observable change.

Fair enough, let's merge this one and I'll make a followup to unbreak the module system.

PR #186, opened 2020-03-31
Let code wrap in tables.

rkirsling:
Replacement for https://github.com/tc39/ecma262/pull/1844. Resolves https://github.com/tc39/ecma262/issues/1843.

To repeat my earlier argument here for good measure:
> [This] should only apply when we've actually run out of width to work with, and has the nice benefit of being responsive if your window is a bit too narrow.

We don't actually need to restrict this to tables-_in-notes_ but it currently suffices to do so.


michaelficarra:
I don't see why we should restrict it to notes. It may save us from having to address this again in the future, and it doesn't hurt the rendering currently, right?

rkirsling:
@michaelficarra I agree—@ljharb, would you be okay with this?

ljharb:
On premise it seems fine; it'd be good to see if there's any actual rendering differences.

rkirsling:
Looking through all of the tables currently in ECMA-262, the only one that's _not_ in a note that's affected by this is Table 42, and just when responding to a restricted viewport width:

image

This feels pretty positive to me.

PR #187, opened 2020-04-01
meta: update URL and repo org

ryzokuken:
The title says it all. /cc @bakkot


bakkot:
Should update [the spec](https://github.com/tc39/ecmarkup/blob/master/spec/index.html) too.

PR #190, opened 2020-04-11
lint: forbid unused variables and trailing whitespace

bakkot:


bakkot:
Though actually instead of adding the lint rules piecemeal we should just adopt prettier.

ljharb:
prettier doesn’t handle no unused vars; we’ll still want to add rules piecemeal i think

bakkot:
> prettier doesn’t handle no unused vars; we’ll still want to add rules piecemeal i think

Yeah, I should have specified just the formatting ones.

PR #191, opened 2020-04-11
Adopt prettier

bakkot:
Let's join the modern world, shall we?


bakkot:
I should mention that I'm not necessarily attached to Prettier here; I just want a consistent, common style which can be applied 100% automatically. Prettier fits the bill. A careful selection of eslint rules which have fixers would work just as well, but it's a lot more work to curate.

ljharb:
luckily the airbnb config already exists :-p but not yet for typescript, so, this is probably the easiest path

.eslintrc.json
```diff
@@ -36,22 +38,13 @@
     }
   ],
   "rules": {
-    "no-extra-parens": [
-      "error",
-      "functions"
-    ],
-    "no-inner-declarations": 0,
-    "consistent-return": 0,
+    "prettier/prettier": "error",
```
ljharb, line 41: the prettier settings (ignore and config) should ideally be specified here, so that prettier is only run via eslint.

package.json
```diff
@@ -11,6 +11,7 @@
 		"pretest": "npm run build",
 		"test": "mocha && npm run lint",
 		"lint": "eslint --ext .js,.ts .",
+		"format": "prettier --write .",
```
ljharb, line 14: similarly, this should be `eslint --fix .`, rather than using prettier to do it

bakkot:
> luckily the airbnb config already exists

Yeah but it enforces a lot of much more controversial non-formatting stuff, and cannot be applied 100% automatically. (Unless there's a version of it which is just formatting? I'd be fine with that.)

.eslintrc.json
bakkot: I don't think that's doable. Also, why do you want to enforce that it's only run via eslint? It shouldn't matter how someone wants to run prettier as long as they use the project's settings.

package.json
bakkot: That exits with an error when there are non-formatting-related linting errors, and so is unsuitable for a command named `format`.

ljharb:
I think we might not be on the same page here :-)

Nobody should ever be running any formatter on a project except the project's own formatter. Running prettier on a project that doesn't explicitly use prettier is not appropriate.

Similarly, eslint is already a linter and formatter combined. It's fine if a project doesn't want to enable non-autofixable rules (altho you miss out on a ton of useful checks that way), but that's a separate choice from "which formatting tool do i want to use".

Using only eslint - which many people have their editors set up to automatically use local settings/versions for - means that we can easily add additional rules in the same single source of truth (the eslint config), without having to maintain two separate configs for linting/formatting.

bakkot:
> Nobody should ever be running any formatter on a project except the project's own formatter. Running prettier on a project that doesn't explicitly use prettier is not appropriate.

Well, sure, but the point of this PR is to make this project use prettier as a formatter. At which point, running it on this project would be appropriate. And it wouldn't matter if you're running it via eslint or not, because it would have the same effect either way.

> Similarly, eslint is already a linter and formatter combined.

It is true that it offers both of those benefits. Here it is serving as a formatter strictly via a plugin which causes it to defer to prettier.

> It's fine if a project doesn't want to enable non-autofixable rules (altho you miss out on a ton of useful checks that way), but that's a separate choice from "which formatting tool do i want to use".

Agreed, and I think non-autofixable non-formatting rules are good. (I have in fact just [added one](https://github.com/tc39/ecmarkup/pull/190).) It is specifically non-autofixable formatting rules which I'd prefer to avoid; I don't think the benefit of marginally more consistent style is worth the overhead of making contributors (including me) manually mess with formatting.

As I said, I'd be fine with replacing this PR with any other method of enforcing a consistent, common style which can be applied 100% automatically. That's the thing I actually want. Prettier does that. Other formatting tools, including eslint unless you do a great deal of work, do not. That's why I've made the choice to use Prettier as the formatting tool.

> Using only eslint - which many people have their editors set up to automatically use local settings/versions for

This enforces prettier _using_ eslint; I'm not sure what issue you see here. (Also, such people tend to also have their editors set up to automatically use prettier as well, in my experience, though it's not necessary if you configure eslint to use prettier as I've done here.)

> means that we can easily add additional rules in the same single source of truth (the eslint config)

Style rules, once established, change very rarely, especially if you use prettier to do styling (since it has so few knobs). So while I agree this is a nice thing to have in principle I think it ought to be given very little weight. Having one place to configure formatting rules and a separate place to configure non-formatting rules seems acceptable to me, in exchange for the benefit of being able to automatically fix all formatting rules.

ljharb:
I think that I can deal with leaving the prettier config in their own places for now (to ensure `prettier --write` works properly), but i'd also rather project formatting be done via `eslint --fix`, and if there's a formatting rule that we don't like, disable it. Does that seem like a reasonable compromise?

bakkot:
> I think that I can deal with leaving the prettier config in their own places for now (to ensure `prettier --write` works properly), but i'd also rather project formatting be done via `eslint --fix`

`eslint --fix` works fine as of this pr. I've just also added a `format` command for convenience which _just_ does formatting. `eslint --fix` would have the effect of doing exactly the same formatting, and then also erroring if you non-formatting issues, so it seems like not the right choice for a command named `format`.

But I guess I could switch it out, or remove it entirely, if you feel particularly strongly about it; I just don't understand why.

> if there's a formatting rule that we don't like, disable it

To be clear, this PR already disables all formatting rules except for "is in the format prettier would output", and adopting prettier generally means sticking with that philosophy going forward.

Issue #173, opened 2020-02-24
Support linting the spec


bakkot:
I would like to start enforcing some basic sanity checks for the spec. Some of them will end up implemented in other projects, but I would like to use this issue to track possible lint rules everywhere.

The main goal of this is to reduce editorial churn and correctness issues, but I know that it would also be helpful to people who build tooling based on the spec, like https://github.com/jscert/jsexplain (so cc @brabalan in case you have any ideas).

Currently these are mostly enforced by editors noticing them or @jmdyck submitting after-the-fact PRs, which isn't sustainable.

Here's a few possibilities to get started. Please suggest more.

- Output is valid HTML
- No trailing whitespace
- In algorithms:
  - Consistent spacing, e.g. `foo ( a, b )` in algorithm headers and `foo(a, b)` in algorithm steps.
  - Parameter names are surrounded by `_`
  - If an `If` has substeps, the `If` line ends with `, then`
  - If an `If` has an `Else` which has substeps, it is spelled `Else`, not `Otherwise`
  - If an `If` has an `else` on the same line, it is spelled `; else ` rather than `, else ` or `; else, ` or `; otherwise ` (we are inconsistent about `else` vs `otherwise` here, and I am OK not enforcing it, but we should at least enforce the semicolon and lack of following space).
  - Lines end in `.`, `:`, `,`, `do`, or `then`
- Grammar lookahead restrictinos and flags are omitted in early error definitions and syntax-directed operations
- In the grammar:
  - Lookahead restrictions and flags do not have spaces between brackets (so ``[lookahead != `let`]`` or `[+Await]` rather than ``[ lookahead != `let` ]`` or `[ +Await ]`, etc)
 - One space before the `:` on the LHS
 - One space between each terminal or nonterminal on the LHS
- A bunch of HTML consistency stuff:
  - tags are lowercase
  - attribute values are quoted (or unquoted; I don't care as long as we're consistent)
  - tags don't have any unexpected attributes ([e.g.](https://github.com/tc39/ecma262/commit/5748238ce37dc438b2985b9a574674a63dc40572))
  - for tags for which the closing tag is optional, they are always included (or not included; again, as long as we're consistent)
  - `≥` is spelled `&ge;` (etc)
  - no unknown `emu` tags
  - consistent indentation
- Consistent British vs American spelling for words where it's an issue, like ["behaviour"](https://github.com/tc39/ecma262/pull/1856)
[...]

PR #199, opened 2020-05-09
basic linting behind --lint-spec

bakkot:
This implements several of the ideas in https://github.com/tc39/ecmarkup/issues/173, and hopefully provides a scaffolding which allows contributing more lint rules relatively easily.

~Depends on https://github.com/tc39/ecmarkdown/pull/69 being released; marked as a draft until then.~ Edit: done, ecmarkdown 5.0.1 is released and used in this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
 	"dependencies": {
 		"bluebird": "^3.7.2",
 		"chalk": "^1.1.3",
-		"ecmarkdown": "/Users/kevin/code/ecmarkdown",
+		"ecmarkdown": "5.0.1",
 		"grammarkdown": "^2.2.0",
 		"highlight.js": "^9.17.1",
 		"html-escape": "^1.0.2",
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
 		"build-release": "tsc -p src",
 		"prepublish": "safe-publish-latest && npm run build-release",
 		"pretest": "npm run build",
-		"test": "mocha && npm run lint",
+		"test": "mocha",
 		"lint": "eslint --ext .js,.ts .",
 		"format": "prettier --write .",
 		"update-pages": "node bin/ecmarkup.js spec/index.html _index.html && git checkout gh-pages && rm index.html && mv _index.html index.html && git add index.html && git commit -m \"update pages\" && git checkout master",
@@ -34,7 +34,7 @@
 	"dependencies": {
 		"bluebird": "^3.7.2",
 		"chalk": "^1.1.3",
-		"ecmarkdown": "NEXT",
+		"ecmarkdown": "/Users/kevin/code/ecmarkdown",
 		"grammarkdown": "^2.2.0",
 		"highlight.js": "^9.17.1",
 		"html-escape": "^1.0.2",
--- a/package.json
+++ b/package.json
@@ -34,8 +34,8 @@
 	"dependencies": {
 		"bluebird": "^3.7.2",
 		"chalk": "^1.1.3",
-		"ecmarkdown": "4.0.0",
-		"grammarkdown": "^2.1.2",
+		"ecmarkdown": "NEXT",
+		"grammarkdown": "^2.2.0",
 		"highlight.js": "^9.17.1",
 		"html-escape": "^1.0.2",
 		"js-yaml": "^3.13.1",
--- a/src/lint/lint.ts
+++ b/src/lint/lint.ts
@@ -22,6 +22,7 @@ import {
   getProductions,
   rhsMatches,
 } from './utils';
+import { collectNodes } from './collect-nodes';
 import lintAlgorithmLineEndings from './rules/algorithm-line-endings';
 
 function composeObservers(...observers: Observer[]): Observer {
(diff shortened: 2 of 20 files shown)


package.json
```diff
@@ -34,8 +34,8 @@
 	"dependencies": {
 		"bluebird": "^3.7.2",
 		"chalk": "^1.1.3",
-		"ecmarkdown": "4.0.0",
-		"grammarkdown": "^2.1.2",
+		"ecmarkdown": "NEXT",
```
ljharb, line 37: why is the tag in all caps? it'll work but it's pretty unconventional

bakkot: Oh, the point is that I'm depending on an unpublished version, and I'll update this to the real version when it's published. That's not an actual tag.

src/lint/lint.ts
```diff
@@ -0,0 +1,400 @@
...
+Currently this checks
+- grammarkdown's built-in sanity checks
+- the productions in the definition of each early error and SDO are defined in the main grammar
+- those productions do not include `[no LineTerminator here]` restrictions or `[+flag]` gating
+- the algorithm linting rules imported above
+
+There's more to do:
+https://github.com/tc39/ecmarkup/issues/173
+*/
+export function lint(
```
michaelficarra, line 51: Can we break this down into smaller functions and have this function just coordinate the linting results?

src/lint/lint.ts
```diff
@@ -0,0 +1,400 @@
...
+          location.startTag.col + (location.startTag.endOffset - location.startTag.startOffset);
+      } else {
+        trueCol += 1;
+      }
+      lintingErrors.push({ line: trueLine, column: trueCol, message });
+    };
+
+    let observer = composeObservers(...algorithmRules.map(f => f(reporter, element)));
+    let tree = parseAlgorithm(
+      sourceText.slice(location.startTag.endOffset, location.endTag.startOffset),
```
michaelficarra, line 359: Can't we just use like `element.innerHTML` here or something?

src/lint/lint.ts
```diff
@@ -0,0 +1,400 @@
...
+      message: string;
+    }) => {
+      let trueLine = location.startTag.line + line - 1; // both jsdom and ecmarkdown have 1-based line numbers, so if we just add we are off by one
+      let trueCol = column;
+      if (line === 0) {
+        trueCol +=
+          location.startTag.col + (location.startTag.endOffset - location.startTag.startOffset);
+      } else {
+        trueCol += 1;
+      }
```
michaelficarra, line 353: This reads better to me.

```suggestion
      let trueCol = line === 0
        ? location.startTag.col + (location.startTag.endOffset - location.startTag.startOffset) + column
        : column + 1;
```

I'm also a little concerned about adding 1 only in the case that it's not the first line. Why is that? Also, I thought the comment above said the lines are 1-based, but we're checking for line `0`?

src/lint/rules/algorithm-line-endings.ts
```diff
@@ -0,0 +1,203 @@
+import type { Node as EcmarkdownNode, Observer } from 'ecmarkdown';
+import type { Reporter } from '../algorithm-error-reporter-type';
+
+export default function (report: Reporter, node: Element): Observer {
```
michaelficarra, line 4: You should summarise the behaviour of this rule in a comment here or in an adjacent markdown document.

src/lint/rules/algorithm-line-endings.ts
```diff
@@ -0,0 +1,203 @@
...
+              line: node.location!.start.line,
+              column: node.location!.start.column,
+              message: `expected line with substeps to end with ":" (found "${last.contents}")`,
+            });
+          }
+        } else if (!/(?:\.|\.\))$/.test(last.contents)) {
+          report({
+            line: node.location!.start.line,
+            column: node.location!.start.column,
+            message: `expected line to end with "." (found "${last.contents}")`,
```
michaelficarra, line 196: This message kinda implies that the only way to resolve this error is to add a `.` at the end. Any way we could say that this is conditional on the other characteristics of the line?

src/lint/rules/algorithm-line-endings.ts
```diff
@@ -0,0 +1,203 @@
...
+      let hasSubsteps = node.sublist !== null;
+
+      if (/^(?:If |Else if)/.test(initialText)) {
+        if (hasSubsteps) {
+          if (node.sublist!.name === 'ol') {
+            if (!/, then$/.test(last.contents)) {
+              report({
+                line: node.location!.start.line,
+                column: node.location!.start.column,
+                message: `expected "If" with substeps to end with ", then" (found "${last.contents}")`,
```
michaelficarra, line 98: I kinda prefer quoting these things with `JSON.stringify` to avoid surprising messages.

src/lint/lint.ts
```diff
@@ -0,0 +1,400 @@
...
+          navigator.moveToAncestor(node => node.kind === SyntaxKind.Production);
+          let nodeName = (navigator.getNode() as Production).name.text!;
+          let paramName = param.name.text!;
+          if (!unusedParameterErrors.has(nodeName)) {
+            unusedParameterErrors.set(nodeName, new Map());
+          }
+          let paramToError = unusedParameterErrors.get(nodeName)!;
+          if (paramToError.has(paramName)) {
+            throw new Error(
+              'duplicate warning for unused parameter ' + paramName + ' of ' + nodeName
```
michaelficarra, line 222: Prefer templates.

test/lint.js
```diff
@@ -0,0 +1,145 @@
...
+describe('linting', function () {
+  describe('grammar sanity', function () {
+    it('unused parameters', async function () {
+      await assertLint(
+        `
+          <emu-grammar type="definition">
+            Statement[${M}a]: \`;\`
+          </emu-grammar>
+        `,
+        "Parameter 'a' is unused."
```
michaelficarra, line 21: Are these messages being passed through from Grammarkdown's built-in linting? I don't like how its capitalisation and quoting are inconsistent with ours. And I prefer that ours does not change because it's better.

test/lint.js
```diff
@@ -0,0 +1,145 @@
...
+let emu = require('../lib/ecmarkup');
+
+function alg([before, after], marker) {
+  return `<emu-alg>${before}${marker}${after}</emu-alg>`;
+}
+
+let M = '@@@'; // Marks the location of the error
+
+describe('linting', function () {
+  describe('grammar sanity', function () {
```
michaelficarra, line 13:
```suggestion
  describe('grammar validity', function () {
```

may be a better choice of words.

test/lint.js
```diff
@@ -0,0 +1,145 @@
+'use strict';
+
+let assert = require('assert');
+let emu = require('../lib/ecmarkup');
+
+function alg([before, after], marker) {
+  return `<emu-alg>${before}${marker}${after}</emu-alg>`;
```
michaelficarra, line 7: I don't see why we actually include the marker. We just need to track positions here, right?

test/lint.js
```diff
@@ -0,0 +1,145 @@
...
+          <emu-alg>
+            1. Return Foo.
+          </emu-alg>
+        `,
+        'Could not find a production matching RHS in syntax-directed operation'
+      );
+    });
+  });
+
+  describe('algorithm line endings', function () {
```
michaelficarra, line 76: This doesn't even come close to covering this rule. Also, I'd probably split it into its own file. And no negative tests?

src/lint/lint.ts
bakkot: `.innerHTML` is not guaranteed to give you the raw source - for example, it will translate `&#97;` in the original source into `a` in the `innerHTML`. Since we care about location information, that difference is unacceptable.

test/lint.js
bakkot: Alas, they are.

src/lint/rules/algorithm-line-endings.ts
ljharb: or https://npmjs.com/object-inspect, if the values might not be JSON-serializable

src/lint/lint.ts
bakkot: I am not convinced this is an improvement.

src/lint/lint.ts
bakkot: > I'm also a little concerned about adding 1 only in the case that it's not the first line. Why is that?

Because jsdom's columns are 1-based, so if we include `location.startTag.col` we have already included the 1.

> Also, I thought the comment above said the lines are 1-based, but we're checking for line `0`?

Nice catch, fixed (and tested).

Also added more comprehensive comments to the places we do location adjustments. The four tools we're using have three different behaviors:

- grammarkdown is 0-based lines and 0-based columns
- ecmarkdown is 1-based lines and 0-based columns (which we did to match eslint)
- jsdom and this linter are 1-based for both (to match editors)

src/lint/rules/algorithm-line-endings.ts
bakkot: Changed it to say "expected freeform line...". I don't have any better ideas.

src/lint/rules/algorithm-line-endings.ts
bakkot: > I kinda prefer quoting these things with JSON.stringify to avoid surprising messages.

Done.

> if the values might not be JSON-serializable

They're all strings.

test/lint.js
bakkot: We could thread location information through, but this is easier and I am not inclined to do the fancier thing just for the tests.

test/lint.js
bakkot: > This doesn't even come close to covering this rule.

Complete tests for this linting rule did not seem necessary to me. Feel free to push up a commit with some, if you want.

> Also, I'd probably split it into its own file.

... Why? That might make sense once we get more tests, I guess, but I don't see the point.

> And no negative tests?

Added a negative test.

test/lint.js
bakkot: Done.

src/lint/rules/algorithm-line-endings.ts
bakkot: Done.

src/lint/lint.ts
bakkot: Done, though.

package.json
```diff
@@ -34,7 +34,7 @@
 	"dependencies": {
 		"bluebird": "^3.7.2",
 		"chalk": "^1.1.3",
-		"ecmarkdown": "NEXT",
+		"ecmarkdown": "/Users/kevin/code/ecmarkdown",
```
ljharb, line 37: ftr you can use a normal version here, and use `npm link` locally, without having to put anything odd in the branch

bakkot: The commit was accidental, but avoiding `npm link` was intentional; at least half the time I try to use it it ends up corrupting the state of my `node_modules`.

ljharb: `rm -rf node_modules && npm install` is a pretty easy fix for any node_modules issues, but k

bakkot: Sure but just not having to fix your node_modules is better

test/lint.js
michaelficarra: Yeah but this strategy won't work for all algorithm texts and marker positions. I'd prefer we do the right thing now so that later PRs don't have to both change up this testing strategy and implement a linting rule. I can pair with you on it if you want, it shouldn't take long.

test/lint.js
michaelficarra: I'm okay with not getting 100% coverage for now, but would like to get to that eventually. We should still split it out of this file though, as we expect to add more linting rules and don't want them to continuously grow this file.

test/lint.js
michaelficarra: Should we prefix grammarkdown messages or somehow indicate they came from grammarkdown?

test/lint.js
bakkot: It's not hard, I just don't think it's worth doing. Feel free to send a followup PR.

test/lint.js
bakkot: Ehhh, that doesn't seem like a distinction users are likely to care about.

test/lint.js
bakkot: > We should still split it out of this file

Done.

test/lint.js
michaelficarra: I did it.

PR #201, opened 2020-05-18
support accepting eslint-compatible formatters; default to codeframe

michaelficarra:


package.json
```diff
@@ -33,8 +33,8 @@
 	"license": "MIT",
 	"dependencies": {
 		"bluebird": "^3.7.2",
-		"chalk": "^1.1.3",
```
bakkot, line 36: This is used elsewhere and can't be removed.

package.json
```diff
@@ -35,6 +35,7 @@
 		"bluebird": "^3.7.2",
 		"chalk": "^1.1.3",
 		"ecmarkdown": "5.0.1",
+		"eslint": "^6.8.0",
```
ljharb, line 38: eslint 7 is out, any reason to stick to v6?

michaelficarra: Because we were already using eslint 6.x as a dev dependency, and I have no interest in dealing with the consequences of bumping that.

michaelficarra:
@ljharb What kind of output do you expect from the CLI then? Also, remember that this is exactly how it works for users of eslint. I think this would be the expected usage.

ljharb:
If we're expecting ecmarkup to work like eslint, it should be an eslint plugin.

If it's a separate tool, it should be pipeable, like other unixy tools.

michaelficarra:
@ljharb I'm really confused. Ecmarkup is a tool that (among other things) produces linting errors. Eslint is also a tool that produces linting errors. There's node modules that take data in that structure and output it for human consumption. eslint has some of them built-in for convenience. This PR causes ecmarkup to conform to the eslint linting error format so that it can use the same formatters that format linting errors produced by eslint.

Am I missing something?

ljharb:
What tools produce eslint-compatible linting errors that aren't explicitly part of the eslint ecosystem? That is very much not a pattern I have ever seen before.

michaelficarra:
@ljharb What is your recommendation? Have the ecmarkup CLI output JSON and pipe the JSON to the eslint formatters, who don't even provide a CLI?

ljharb:
My recommendation would be that ecmarkup primarily produce output for humans; and that we create an eslint plugin that incorporates ecmarkup to provide linter warnings as part of eslint. It makes no sense to me to use an eslint formatter on "not the output of `eslint`".

bakkot:
eslint formatters are things which print linting errors. We want to print linting errors. It seems reasonable to use eslint formatters to do so.

src/lint/collect-grammar-diagnostics.ts
```diff
@@ -97,6 +104,7 @@ export function collectGrammarDiagnostics(
   // *******************
   // Check that SDOs and Early Errors are defined in terms of productions which actually exist
   // Also filter out any "unused parameter" warnings for grammar productions for which the parameter is used in an early error or SDO
+  const ruleId = 'undefined-nonterminal';
```
bakkot, line 107: There's actually four distinct errors in this section:

- when the LHS has no corresponding definition
- when the LHS exists but none of its productions match the RHS given in the SDO
- when the RHS contains a "[no LineTerminator here]" restriction
- when the production on the RHS is gated on a parameter (as in `[+a] Whatever`)

They should probably have different IDs. (They already have different messages.)

ljharb:
There's no TAP-like standard for linting errors; they don't print linting errors, they format *eslint errors* - errors generated by eslint.

bakkot:
Regardless of what you choose to call the kinds of errors they print they work perfectly well for this purpose.

bakkot:
@michaelficarra it looks like I accidentally disabled the enforcement of linting in this project; would you re-enable it and fix your resulting linting errors?

package.json
```diff
@@ -8,7 +8,6 @@
 		"build": "tsc -p src -sourceMap",
 		"build-release": "tsc -p src",
 		"prepublish": "safe-publish-latest && npm run build-release",
-		"pretest": "npm run build",
```
ljharb, line 11: why remove this, and move it to travis.yml? it's usually important that `npm test` locally runs as much as possible.

src/cli.ts
```diff
@@ -30,20 +29,33 @@ const build = debounce(async function build() {
...
+      let formatter = require(descriptor);
+
+      opts.reportLintErrors = (errors: LintingError[], sourceText: string) => {
+        if (errors.length < 1) return;
+
+        let results = [
+          {
+            filePath: args.infile,
+            // for now, everything is an error (2 in eslint terms)
+            messages: errors.map(e => Object.assign({ severity: 2 }, e)),
```
ljharb, line 47:
```suggestion
            messages: errors.map(e => ({ severity: 2, ...e })),
```

package.json
michaelficarra: It's so I can iterate on tests without having to run the build every time. With this, there's no way to just run the tests.

ljharb: Ah, usually i achieve that by putting the actual tests in a `tests-only` script, and then have `"test": "npm run tests-only"`

michaelficarra: That seems a little backwards. Why not have the thing called "test" run the tests, the thing called "build" run the build, the thing called "lint" run the lint? If we want convenient combined ones we can make them, like "build-and-test" or something, but it's also pretty easy to run `npm run build && npm test`.

ljharb: Because the decade-old node module convention is that `npm test` runs everything.

src/lint/collect-grammar-diagnostics.ts
michaelficarra: Changed rule IDs for the latter two, but there were no tests to update since these were untested.

package.json
bakkot: Really? That's not what [babel](https://github.com/babel/babel/blob/73ab27c11a88f9bbc33caec6e8b781b8c3b7602c/Makefile#L148), [prettier](https://github.com/prettier/prettier/blob/765e3b40251abbf7f636580dfcb99f69da3dd42d/package.json#L126), [react](https://github.com/facebook/react/blob/5aa967b69b255ede89c03dda8cd617b61c492cdb/package.json#L110), [vue](https://github.com/vuejs/vue/blob/0baa129d4cad44cf1847b0eaf07e95d4c71ab494/package.json#L29), etc do. Whence this certainty about the convention?

ljharb: As always, the large projects often defy long tail conventions; conventions are set by the ecosystem, not by the major players.

However, in your babel link, specifically, they do precisely what i say - `npm test` runs `make test` which runs both `lint` and `test-only`.

bakkot: > As always, the large projects often defy long tail conventions; conventions are set by the ecosystem, not by the major players.

OK. If I find a [random list of JS projects](https://github.com/sorrycc/awesome-javascript), scroll down to the utilities, and start looking for things which a.) are not the "large projects" and b.) have both a `test` and a `build`, I find [a](https://github.com/rtfpessoa/diff2html/blob/d7d08f0c3687d02a841feb37035046bb532acd89/package.json#L54) [bunch](https://github.com/MrRio/jsPDF/blob/2c5e6c7cb0ccaaabdbf7a823cf5bb076b57ef9b0/package.json#L88) [of](https://github.com/cujojs/most/blob/332137c3878fa199cd6b67781f3cbc13f10e7149/package.json#L17) [projects](https://github.com/swannodette/mori/blob/46c9194b3c4bc93fc4b402925b9417add3b0a3ba/package.json#L48) [which](https://github.com/iamkun/dayjs/blob/9c0523e4dc0fa3c1ae59065e48a18b606e708d28/package.json#L9) [split](https://github.com/panzerdp/voca/blob/d3183a8a7099dae326ae4c6b4eeeab74e2cf4a64/package.json#L42) [them](https://github.com/mathiasbynens/he/blob/36afe179392226cf1b6ccdb16ebbb7a5a844d93a/package.json#L39) and exactly [one](https://github.com/taylorhakes/fecha/blob/91680e4db1415fea33eac878cfd889c80a7b55c7/package.json#L9) where `test` runs `build`.

Again, whence this certainty about the convention?


> they do precisely what i say

In babel `make test` specifically does not run `build`.

michaelficarra: I think it's clear at this point that the convention is not universal, even if common as @ljharb claims. I have seen enough examples to convince me that it's okay to violate this convention, if it does indeed exist. In that case, we should do what is best for our desired development workflow, and I think that's keeping them separate.

ljharb: ¯\\\_(ツ)_/¯ in my experience, not following this convention often causes people to miss non-test issues until they open up a PR, causing review noise and inconvenience for them (it happens a lot on the bigger projects too, it just gets lost in the noise they already have to deal with). we can go with this until it inevitably becomes a problem, sure.

PR #203, opened 2020-05-19
use files array instead of npmignore

bakkot:
I hadn't been updating `.npmignore`, which meant we were including dotfiles like `.eslintignore`, which was silly.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,9 +1,0 @@
-.vscode
-.editorconfig
-.eslintrc
-.gitignore
-.travis.yml
-lib/**/*.js.map
-src
-test
-deploy.sh
--- a/package.json
+++ b/package.json
@@ -17,6 +17,14 @@
 		"postversion": "auto-changelog && git add CHANGELOG.md && git commit --no-edit --amend && git tag -f \"v$(node -e \"console.log(require('./package.json').version)\")\""
 	},
 	"bin": "bin/ecmarkup.js",
+	"files": [
+		"/bin",
+		"/lib",
+		"/js",
+		"/css",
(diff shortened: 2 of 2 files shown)


ljharb:
"files" is dangerous, and i'd prefer to revert this; accidentally including dotfiles is a zero-damage minor annoyance; accidentally not including production files breaks consumers.

michaelficarra:
@ljharb Then presumably we'd fix it and create a new patch release? Also, how likely is that to happen? We're including whole directories, that kind of structure doesn't change often, if ever.

ljharb:
In that case yes, we'd have to fix it and create a new patch (just like if we discover files are accidentally included). I agree it's very unlikely, and in this case, our only consumer is ecma262 which has a lockfile, but between in general a nonzero chance of breakage, versus a zero chance of breakage and a nonzero chance of minor annoyance, it just doesn't make any sense to choose the nonzero damage tradeoff.

michaelficarra:
I prefer loud errors that can be resolved quickly over quiet ones that go unnoticed forever.

bakkot:
@ljharb Using `.npmignore` does not make it impossible that we fail to include a necessary file, since there is nothing which prevents us from adding a necessary file which matches a `.npmignore` pattern.

If we are worried we might not include some necessary file we should test that explicitly. I'll write a quick PR.

ljharb:
@michaelficarra extra files are not errors, they're no-ops. It's nothing more than an optimization to exclude them.

@bakkot you're totally right that testing what `npm pack` picks up resolves this issue with either approach; i'll take a look at the PR.

PR #205, opened 2020-05-20
add lint for header format

bakkot:
cf https://github.com/tc39/ecma262/pull/1998/files#diff-3540caefa502006d8a33cb1385720803R37872

### DIFF
--- a/src/lint/collect-header-diagnostics.ts
+++ b/src/lint/collect-header-diagnostics.ts
@@ -1,0 +1,104 @@
+import type { LintingError } from './algorithm-error-reporter-type';
+
+import { getLocation } from './utils';
+
+const ruleId = 'header-format';
+
+export function collectHeaderDiagnostics(
+  dom: any,
+  headers: { element: Element; contents: string }[]
+) {
+  let lintingErrors: LintingError[] = [];
+
+  for (let { element, contents } of headers) {
+    if (!/\(.*\)$/.test(contents) || / Operator \( `[^`]+` \)$/.test(contents)) {
+      continue;
+    }
+
+    function indexToLineAndColumn(index: number) {
+      let headerLines = contents.split('\n');
+      let headerLine = 0;
+      let seen = 0;
+      while (true) {
+        if (seen + headerLines[headerLine].length >= index) {
+          break;
+        }
+        seen += headerLines[headerLine].length + 1; // +1 for the '\n'
+        ++headerLine;
+      }
+      let headerColumn = index - seen;
+
+      let elementLoc = getLocation(dom, element);
+      let line = elementLoc.startTag.line + headerLine;
+      let column =
+        headerLine === 0
+          ? elementLoc.startTag.col +
+            (elementLoc.startTag.endOffset - elementLoc.startTag.startOffset) +
+            headerColumn
+          : headerColumn + 1;
+
+      return { line, column };
+    }
+
+    let name = contents.substring(0, contents.indexOf('('));
+    let params = contents.substring(contents.indexOf('(') + 1, contents.length - 1);
+
+    if (/ $/.test(name)) {
+      name = name.substring(0, name.length - 1);
+    } else {
+      let { line, column } = indexToLineAndColumn(name.length);
+      lintingErrors.push({
+        ruleId,
+        nodeType: 'H1',
+        line,
+        column,
+        message: 'expected header to have a space before the argument list',
+      });
+    }
+
+    let nameMatches = [
+      /^(Runtime|Static) Semantics: [A-Z][A-Za-z0-9/]*$/,
+      /^(Number|BigInt)::[a-z][A-Za-z0-9]*$/,
+      /^\[\[[A-Z][A-Za-z0-9]*\]\]$/,
+      /^_[A-Z][A-Za-z0-9]*_$/,
+      /^[A-Za-z][A-Za-z0-9]*(\.[A-Za-z][A-Za-z0-9]*)*( \[ @@[a-z][a-zA-Z]+ \])?$/,
+      /^%[A-Z][A-Za-z0-9]*%(\.[A-Za-z][A-Za-z0-9]*)*( \[ @@[a-z][a-zA-Z]+ \])?$/,
+    ].some(r => r.test(name));
+
+    if (!nameMatches) {
+      let { line, column } = indexToLineAndColumn(0);
+      lintingErrors.push({
+        ruleId,
+        nodeType: 'H1',
+        line,
+        column,
+        message: `expected operation to have a name like 'Example', 'Runtime Semantics: Foo', 'Example.prop', etc, but found ${JSON.stringify(
+          name
+        )}`,
+      });
+    }
+
+    let paramsMatches =
+      params.match(/\[/g)?.length === params.match(/\]/g)?.length &&
+      [
+        /^ $/,
+        /^ \. \. \. $/,
+        /^ (_[A-Za-z0-9]+_, )*\.\.\._[A-Za-z0-9]+_ $/,
+        /^ (_[A-Za-z0-9]+_, )*… (, _[A-Za-z0-9]+_)+ $/,
+        /^ (\[ )?_[A-Za-z0-9]+_(, _[A-Za-z0-9]+_)*( \[ , _[A-Za-z0-9]+_(, _[A-Za-z0-9]+_)*)*( \])* $/,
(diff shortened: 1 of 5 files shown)


src/lint/collect-header-diagnostics.ts
```diff
@@ -0,0 +1,104 @@
...
+  headers: { element: Element; contents: string }[]
+) {
+  let lintingErrors: LintingError[] = [];
+
+  for (let { element, contents } of headers) {
+    if (!/\(.*\)$/.test(contents) || / Operator \( `[^`]+` \)$/.test(contents)) {
+      continue;
+    }
+
+    function indexToLineAndColumn(index: number) {
```
michaelficarra, line 18: I'd prefer this be a pure function of the text content, the index into the text content, and the starting { line, column, offset } of the text content.

src/lint/collect-header-diagnostics.ts
```diff
@@ -0,0 +1,104 @@
...
+            headerColumn
+          : headerColumn + 1;
+
+      return { line, column };
+    }
+
+    let name = contents.substring(0, contents.indexOf('('));
+    let params = contents.substring(contents.indexOf('(') + 1, contents.length - 1);
+
+    if (/ $/.test(name)) {
```
michaelficarra, line 46:
```suggestion
    if (/[^\s] $/.test(name)) {
```

This way we can get an error about spacing if there's more than one space instead of an error about naming. We can amend the message to say exactly one.

src/lint/collect-header-diagnostics.ts
```diff
@@ -0,0 +1,104 @@
...
+        nodeType: 'H1',
+        line,
+        column,
+        message: 'expected header to have a space before the argument list',
+      });
+    }
+
+    let nameMatches = [
+      /^(Runtime|Static) Semantics: [A-Z][A-Za-z0-9/]*$/,
+      /^(Number|BigInt)::[a-z][A-Za-z0-9]*$/,
```
michaelficarra, line 61: I don't think we need to restrict this to `Number`/`BigInt`.

```suggestion
      /^[A-Z][A-Za-z0-9]*::[a-z][A-Za-z0-9]*$/,
```

src/lint/collect-header-diagnostics.ts
```diff
@@ -0,0 +1,104 @@
...
+      /^\[\[[A-Z][A-Za-z0-9]*\]\]$/,
+      /^_[A-Z][A-Za-z0-9]*_$/,
+      /^[A-Za-z][A-Za-z0-9]*(\.[A-Za-z][A-Za-z0-9]*)*( \[ @@[a-z][a-zA-Z]+ \])?$/,
```
michaelficarra, line 64: I think each of these deserve a commend above with an example of what they're matching just so someone can quickly find the relevant regexp.

src/lint/collect-header-diagnostics.ts
```diff
@@ -0,0 +1,104 @@
...
+    let name = contents.substring(0, contents.indexOf('('));
+    let params = contents.substring(contents.indexOf('(') + 1, contents.length - 1);
+
+    if (/ $/.test(name)) {
+      name = name.substring(0, name.length - 1);
+    } else {
+      let { line, column } = indexToLineAndColumn(name.length);
+      lintingErrors.push({
+        ruleId,
+        nodeType: 'H1',
```
michaelficarra, line 52:
```suggestion
        nodeType: element.tagName,
```

src/lint/collect-header-diagnostics.ts
bakkot: That splits the logic in an awkward way. I've changed it to be pure and take an element location, though.

bakkot:
@michaelficarra addressed comments.

src/lint/collect-header-diagnostics.ts
```diff
@@ -0,0 +1,113 @@
...
+        /^ $/,
+
+        // Object ( . . . )
+        /^ \. \. \. $/,
+
+        // String.raw ( _template_, ..._substitutions_ )
+        /^ (_[A-Za-z0-9]+_, )*\.\.\._[A-Za-z0-9]+_ $/,
+
+        // Function ( _p1_, _p2_, &hellip; , _pn_, _body_ )
+        /^ (_[A-Za-z0-9]+_, )*… (, _[A-Za-z0-9]+_)+ $/,
```
michaelficarra, line 89:
```suggestion
        /^ (_[A-Za-z0-9]+_, )*…( , _[A-Za-z0-9]+_)* $/,
```

src/lint/collect-header-diagnostics.ts
```diff
@@ -0,0 +1,113 @@
...
+
+      // [[GetOwnProperty]]
+      /^\[\[[A-Z][A-Za-z0-9]*\]\]\s*$/,
+
+      // _NativeError_
+      /^_[A-Z][A-Za-z0-9]*_\s*$/,
+
+      // CreateForInIterator
+      // Object.fromEntries
+      // Array.prototype [ @@iterator ]
```
michaelficarra, line 51: We should support `Array.prototype [ %Symbol.iterator% ]` as well, since we're moving to that form.

src/lint/collect-header-diagnostics.ts
bakkot: I explicitly do not want to allow the form which has `…` without a following argument.

michaelficarra: What about a preceding argument?

```suggestion
        /^ (_[A-Za-z0-9]+_, )+…( , _[A-Za-z0-9]+_)+ $/,
```

At the very least, you need this one:

```suggestion
        /^ (_[A-Za-z0-9]+_, )*…( , _[A-Za-z0-9]+_)+ $/,
```

src/lint/collect-header-diagnostics.ts
bakkot: I'd prefer that we disallow it now and make it legal as part of the PR which changes the format, so we are never in a position where both are accepted.

src/lint/collect-header-diagnostics.ts
bakkot: > What about a preceding argument?

I'm fine with this matching `(..., _a_)`.

>  At the very least, you need this one:

No I don't. That would match `… , _a_ , _b_` instead of `… , _a_, _b_`. I want the second of those.

michaelficarra: Oops, I meant

```suggestion
        /^ (_[A-Za-z0-9]+_, )*…(, _[A-Za-z0-9]+_)+ $/,
```

src/lint/collect-header-diagnostics.ts
michaelficarra: But that's in a different repo. There's a currently open PR. How do you expect that to work?

src/lint/collect-header-diagnostics.ts
bakkot: No, that comma is [there upstream](https://tc39.es/ecma262/#sec-function-p1-p2-pn-body), and I'm just trying to match the current upstream conventions.

src/lint/collect-header-diagnostics.ts
bakkot: We'd change it here, bump it, release it, and include the change in the upstream PR as part of the process of merging that PR.

michaelficarra: I don't love it but I'll leave it to @ljharb to fight this fight.

src/lint/collect-header-diagnostics.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    let name = contents.substring(0, contents.indexOf('('));
+    let params = contents.substring(contents.indexOf('(') + 1, contents.length - 1);
+
+    if (!/[\S] $/.test(name)) {
+      let { line, column } = indexWithinElementToTrueLocation(
+        getLocation(dom, element),
+        contents,
+        name.length
```
michaelficarra, line 25:
```suggestion
        name.length - 1,
```

bakkot: Hm. It feels kind of weird to point to the `e` in `Example( )`, but sure, I guess.

michaelficarra: It will also point to the second space in <pre><code>Example  ( )</pre></code> though.

PR #207, opened 2020-05-24
lint: enforce some spelling conventions

bakkot:
This rule is tripped 29 times on ECMA-262 master: in 26 places where we say `*this* object` instead of (as in the other 429 uses of `*this*`) `*this*` value", in 1 place where we say `behavior` instead of (as in the other 132 places) `behaviour`, and in 2 places where we say `the empty string` instead of (as in the other 76 places) `the empty String`.

I'll fix those as part of upstreaming this.

### DIFF
--- a/src/lint/collect-spelling-diagnostics.ts
+++ b/src/lint/collect-spelling-diagnostics.ts
@@ -8,7 +8,7 @@ let ruleId = 'spelling';
 let matchers = [
   {
     pattern: /\*this\* object/giu,
-    message: 'Prefer "the *this* value"',
+    message: 'Prefer "*this* value"',
   },
   {
     pattern: /1's complement/giu,
@@ -24,7 +24,11 @@ let matchers = [
   },
   {
     pattern: /behavior/giu,
-    message: 'ECMA-262 uses UK spelling ("behaviour")',
+    message: 'ECMA-262 uses Oxford spelling ("behaviour")',
+  },
+  {
+    pattern: /[Tt]he empty string/gu,
+    message: 'Prefer "the empty String"',
   },
 ];
 
--- a/src/lint/collect-spelling-diagnostics.ts
+++ b/src/lint/collect-spelling-diagnostics.ts
@@ -55,7 +55,7 @@ export function collectSpellingDiagnostics(sourceText: string) {
     }
     if (errors.length === 0) {
       throw new Error(
-        'Ecmarkup has a bug: the spell checker reported an error, but could find one. Please report this at https://github.com/tc39/ecmarkup/issues/new.'
+        'Ecmarkup has a bug: the spell checker reported an error, but could not find one. Please report this at https://github.com/tc39/ecmarkup/issues/new.'
       );
     }
     return errors;
--- a/src/lint/collect-spelling-diagnostics.ts
+++ b/src/lint/collect-spelling-diagnostics.ts
@@ -1,0 +1,60 @@
+import type { LintingError } from './algorithm-error-reporter-type';
+
+import { offsetToLineAndColumn } from './utils';
+
+let ruleId = 'spelling';
+
+// Note that these will be composed, so cannot contain backreferences
+let matchers = [
+  {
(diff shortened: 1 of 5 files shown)


bakkot:
Note that it is plausible that we actually do want to say `*this* object`; @rkirsling [here](https://github.com/tc39/ecma262/pull/1736) observes that it appears to be a shorthand for "this value (which is an object)". It's only ever used to describe the behavior of built-in methods (e.g. [`String.prototype.concat`](https://tc39.es/ecma262/#sec-string.prototype.concat)). But I don't think it makes sense to have that distinction; in those cases nothing enforces that `*this*` is in fact an object.

src/lint/collect-spelling-diagnostics.ts
```diff
@@ -0,0 +1,60 @@
...
+    pattern: /2's complement/giu,
+    message: 'Prefer "two\'s complement"',
+  },
+  {
+    pattern: /\*0\*/gu,
+    message: 'The Number value 0 should be written "*+0*", to unambiguously exclude "*-0*"',
+  },
+  {
+    pattern: /behavior/giu,
+    message: 'ECMA-262 uses UK spelling ("behaviour")',
```
rkirsling, line 27:
```suggestion
    message: 'ECMA-262 uses Oxford spelling ("behaviour")',
```
nit: Probably worth being [precise](https://en.wikipedia.org/wiki/Oxford_spelling).

src/lint/collect-spelling-diagnostics.ts
```diff
@@ -0,0 +1,60 @@
...
+
+import { offsetToLineAndColumn } from './utils';
+
+let ruleId = 'spelling';
+
+// Note that these will be composed, so cannot contain backreferences
+let matchers = [
+  {
+    pattern: /\*this\* object/giu,
+    message: 'Prefer "the *this* value"',
```
rkirsling, line 11: I think the intent here (with `the` in the message but not in the pattern) is clear, so I don't object, but I will note that there are a few instances of `a *this* value` in the spec.

ljharb:
I'd agree; the receiver only has to be an object in sloppy mode, and builtin methods have `RequireObjectCoercible` to account for nullish values, so `this value` seems to be the proper choice.

src/lint/collect-spelling-diagnostics.ts
bakkot: Removed the `the`.

src/lint/collect-spelling-diagnostics.ts
bakkot: Done, thanks.

src/lint/collect-spelling-diagnostics.ts
```diff
@@ -0,0 +1,64 @@
...
+          line,
+          column,
+          message,
+        });
+        match = pattern.exec(sourceText);
+      }
+    }
+    if (errors.length === 0) {
+      throw new Error(
+        'Ecmarkup has a bug: the spell checker reported an error, but could find one. Please report this at https://github.com/tc39/ecmarkup/issues/new.'
```
michaelficarra, line 58: I don't understand this message. Is there a typo?

bakkot: Ah, yeah, it should read "could *not*". Fixed.

src/lint/collect-spelling-diagnostics.ts
```diff
@@ -0,0 +1,64 @@
...
+  {
+    pattern: /1's complement/giu,
+    message: 'Prefer "one\'s complement"',
+  },
+  {
+    pattern: /2's complement/giu,
+    message: 'Prefer "two\'s complement"',
+  },
+  {
+    pattern: /\*0\*/gu,
```
michaelficarra, line 22:
```suggestion
    pattern: /\*0(?:<sub>𝔽</sub>)?\*/gu,
```

bakkot: Eh, doesn't seem worth it.

PR #208, opened 2020-05-27
lint: allow lines to end in <pre> tags

bakkot:
See https://github.com/tc39/ecma262/pull/2013#issuecomment-634169177.

Contains version bump commit (minor because of the inclusion of two new linting rules, in https://github.com/tc39/ecmarkup/pull/205 and https://github.com/tc39/ecmarkup/pull/207), so please **rebase**, not squash.

### DIFF
--- a/test/lint-algorithm-line-endings.js
+++ b/test/lint-algorithm-line-endings.js
@@ -69,6 +69,21 @@ describe('linting algorithms', function () {
       );
     });
 
+    it('pre', async function () {
+      await assertLint(
+        positioned`<emu-alg>
+            1. ${M}Let _constructorText_ be the source text
+            <pre><code class="javascript">constructor() {}</code></pre>
+              1. Foo.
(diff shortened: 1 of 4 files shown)


test/lint-algorithm-line-endings.js
```diff
@@ -93,6 +108,9 @@ describe('linting algorithms', function () {
           1. Other.
           1. Other:
             1. Substep.
+          1. Let _constructorText_ be the source text
+          <pre><code class="javascript">constructor() {}</code></pre>
+          1. Set _constructor_ to ParseText(_constructorText_, |MethodDefinition[~Yield, ~Await]|).
```
michaelficarra, line 113: Was this line needed?

bakkot: It isn't strictly needed. I just wanted to confirm that the PR this change is intended to allow will in fact be allowed by it.

PR #210, opened 2020-05-29
Lint: add some rules about linebreaks

bakkot:
The "No more than one blank line is allowed" rule is tripped 23 times on ECMA-262 master. The other two are not tripped at all.

I'll fix those as part of upstreaming this.


src/lint/collect-spelling-diagnostics.ts
```diff
@@ -34,6 +34,19 @@ let matchers = [
...
     message: 'Trailing spaces are not allowed',
   },
+  {
+    // The negative lookbehind is to get a better match location
+    pattern: /(?<=\n)\n\n+/gu,
+    message: 'No more than one blank line is allowed',
+  },
+  {
+    pattern: /(?<=<emu-clause.*>\n)\n\s*<h1>/giu,
+    message: "There should not be a blank line between a clause's opening tag and its header",
```
ljharb, line 44: what about the last line of a clause, and its closing tag?

bakkot: Sure, added.

src/lint/collect-spelling-diagnostics.ts
```diff
@@ -34,6 +34,23 @@ let matchers = [
     pattern: /[ \t]+\n/gu,
     message: 'Trailing spaces are not allowed',
   },
+  {
+    pattern: /(?<=\n)\n\n+/gu,
```
michaelficarra, line 38:
```suggestion
    pattern: /(?<=(^|[^\n])\n\n)\n+/gu,
```

This does two things:

1. Points to the second empty line, not the first.
1. Only raises a single linting error, even for very long runs of empty lines.

You should add tests confirming these things.

src/lint/collect-spelling-diagnostics.ts
```diff
@@ -34,6 +34,23 @@ let matchers = [
...
+    message: 'No more than one blank line is allowed',
+  },
+  {
+    pattern: /(?<=<emu-clause.*>\n)\n\s*<h1>/giu,
+    message: "There should not be a blank line between a clause's opening tag and its header",
+  },
+  {
+    pattern: /(?<=\n)\n[ \t]*<\/emu-clause>/giu,
```
michaelficarra, line 46:
```suggestion
    pattern: /(?<=(^|[^\n])\n)\n\s*<\/emu-clause>/giu,
```

for the same reasons as above.

src/lint/collect-spelling-diagnostics.ts
bakkot: Done.

src/lint/collect-spelling-diagnostics.ts
bakkot: Done, though I used `/(?<=(^|[^\n])\n)\n+[ \t]*<\/emu-clause>/giu` instead of your suggestion because I don't actually want to match vertical tabs or whatever.

bakkot: Didn't add a test for this one because it's impossible to test without also tripping the two-blank-lines rule, and I didn't really want to bother adding support for multiple linting errors.

PR #212, opened 2020-06-15
lint: add lint for algorithm line numbers

bakkot:
This adds a linting rule which enforces that algorithm steps are all numbered as `1.`, except at the top level of an algorithm whose first item starts with some other number (which is used when providing an alternative definitions of a step which occurs elsewhere).

This rule is tripped 29 times on ECMA-262 master. I'l fix those as part of upstreaming this.

Contains version bump commit, so please **rebase**, not squash.

### DIFF
--- a/src/lint/rules/algorithm-step-numbering.ts
+++ b/src/lint/rules/algorithm-step-numbering.ts
@@ -1,0 +1,50 @@
+import type { Node as EcmarkdownNode, Observer } from 'ecmarkdown';
+import type { LintingError } from '../algorithm-error-reporter-type';
+
+const ruleId = 'algorithm-step-numbering';
+
+/*
+Checks that step numbers are all `1`, with the exception of top-level lists whose first item is not `1`.
+*/
+export default function (
+  report: (e: LintingError) => void,
+  node: Element,
+  algorithmSource: string
+): Observer {
+  const nodeType = node.tagName;
+  let depth = -1;
+  let topLevelIsOne = false;
+  return {
+    enter(node: EcmarkdownNode) {
+      if (node.name === 'ol') {
+        ++depth;
+        if (depth === 0) {
+          topLevelIsOne = node.start === 1;
+        }
+      } else if (node.name === 'ordered-list-item') {
+        if (depth === 0 && !topLevelIsOne) {
+          return;
+        }
+        let itemSource = algorithmSource.slice(
+          node.location!.start.offset,
+          node.location!.end.offset
+        );
+        let match = itemSource.match(/^(\s*)(\d+\.) /)!;
+        if (match[2] !== '1.') {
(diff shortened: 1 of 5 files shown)


src/lint/rules/algorithm-step-numbering.ts
```diff
@@ -0,0 +1,50 @@
...
+Checks that step numbers are all `1`, with the exception of top-level lists whose first item is not `1`.
+*/
+export default function (
+  report: (e: LintingError) => void,
+  node: Element,
+  algorithmSource: string
+): Observer {
+  const nodeType = node.tagName;
+  let depth = -1;
+  let topLevelIsNotOne = false;
```
michaelficarra, line 16: Invert this.

src/lint/rules/algorithm-step-numbering.ts
```diff
@@ -0,0 +1,50 @@
...
+        }
+      } else if (node.name === 'ordered-list-item') {
+        if (depth === 0 && topLevelIsNotOne) {
+          return;
+        }
+        let itemSource = algorithmSource.slice(
+          node.location!.start.offset,
+          node.location!.end.offset
+        );
+        let match = itemSource.match(/^([ \t]*)([^.]+\.) /)!;
```
michaelficarra, line 32:
```suggestion
        let match = itemSource.match(/^(\s*)(\d+\.) /)!;
```

They always start with a number, right? Also, is it possible for the match to fail? Should we account for that?

src/lint/rules/algorithm-step-numbering.ts
```diff
@@ -0,0 +1,50 @@
...
+        if (match[2] !== '1.') {
+          report({
+            ruleId,
+            nodeType,
+            line: node.location!.start.line,
+            column: node.location!.start.column + match[1].length,
+            message: `expected line number to be "1." (found ${JSON.stringify(match[2])})`,
```
michaelficarra, line 39:
```suggestion
            message: `expected step number to be "1." (found ${JSON.stringify(match[2])})`,
```

test/lint-algorithms.js
```diff
@@ -115,4 +116,97 @@ describe('linting algorithms', function () {
       `);
     });
   });
+
+  describe('line numbering', function () {
```
michaelficarra, line 120:
```suggestion
  describe('step numbering', function () {
```

src/lint/rules/algorithm-step-numbering.ts
bakkot: They do always start with a number, and it is not possible for the match to fail.

bakkot:
@michaelficarra Addressed comments. I'll reorder the commits and squash down to just the two before merging once this is approved.

Issue #214, opened 2020-06-15
Obscure error message in linter when missing a closing tag


ptomato:
Forgetting an `</emu-alg>` leads to an error message that gives no clue to what the problem is or where it's located:
```html
<!doctype html>
<meta charset="utf8">

<pre class=metadata>
title: Bug proposal
stage: 1
contributors: E. S. Expert
</pre>

<emu-clause id="sec-bug1">
  <h1>Bug</h1>
  <emu-alg>
    1. Let this be a bug.
  <!--/emu-alg-->
</emu-clause>
```
Stack trace:
```
Error: could not find location
    at Object.getLocation (/.../node_modules/ecmarkup/lib/lint/utils.js:48:15)
    at Object.collectAlgorithmDiagnostics (/.../node_modules/ecmarkup/lib/lint/collect-algorithm-diagnostics.js:38:32)
    at Object.lint (/.../node_modules/ecmarkup/lib/lint/lint.js:24:59)
    at Spec.<anonymous> (/.../node_modules/ecmarkup/lib/Spec.js:163:24)
    at Generator.next (<anonymous>)
    at fulfilled (/.../node_modules/ecmarkup/lib/Spec.js:5:58)
```
Perhaps JSDom has some sort of feature that can verify that all tags are closed, that could be run on the DOM tree as a first stage of the linter?

PR #215, opened 2020-06-15
better errors for unclosed emu-alg/emu-grammar tags

bakkot:
Fixes https://github.com/tc39/ecmarkup/issues/214.

### DIFF
--- a/src/lint/collect-algorithm-diagnostics.ts
+++ b/src/lint/collect-algorithm-diagnostics.ts
@@ -37,7 +37,7 @@ export function collectAlgorithmDiagnostics(
 
     if (location.endTag == null) {
       lintingErrors.push({
-        ruleId: 'no-close-tag',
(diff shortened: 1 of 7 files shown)


src/lint/collect-algorithm-diagnostics.ts
```diff
@@ -35,6 +35,17 @@ export function collectAlgorithmDiagnostics(
     let element = algorithm.element;
     let location = getLocation(dom, element);
 
+    if (location.endTag == null) {
+      lintingErrors.push({
+        ruleId: 'no-close-tag',
```
michaelficarra, line 40: `missing-close-tag` might be less ambiguous. Rules prefixed with `no-` in the eslint world usually imply that it is restricting the use of the thing.

bakkot: Done.

PR #216, opened 2020-06-15
Some fixes for working with <ins> and <del>

ptomato:
I've been trying to use the linter on tc39/proposal-temporal. In some sections we have proposed amendments to the spec, with `<ins>` and `<del>` tags, which trip up the linter. We also have some to-do items marked with `<mark>` which I feel should not trip up the linter either, although that is a bit more debatable I guess; maybe `<mark>` should not be allowed in tc39/ecma262.

Concretely, this pull request makes the following changes:
- If an algorithm line is completely surrounded with an `<ins>`, `<del>`, or `<mark>` tag, the tag is ignored and the line is linted as if the tag were not present.
- `<del>` tags are ignored in header lines.


bakkot:
Thanks, I'll take a look.

A general comment: The linter is not currently designed for use on proposals. This is most glaring in that it enforces some whole-program consistency: e.g., every syntax-directed operation or early error is required to have a corresponding definition in the main grammar. The intention of naming the flag `--lint-spec` was that we could have a different flag, perhaps `--lint-proposal`, which could implement proposal-specific linting.

ptomato:
Got it, thanks. Well, anyway, even running `--lint-spec` on proposal-temporal has been worth it, so I'd continue doing it anyway until something really incompatible happens. Feel free to close this if these changes don't belong in `--lint-spec`, though.

ptomato:
Ping?

bakkot:
Sorry, backlogged on other stuff, I'll get to this when I get a chance.

src/lint/collect-nodes.ts
```diff
@@ -142,3 +142,16 @@ export function collectNodes(
...
   return { success: true, mainGrammar, headers, sdos, earlyErrors, algorithms };
 }
+
+function textContentExcludingDeleted(node: Node): string {
+  let retval = '';
+  node.childNodes.forEach(value => {
+    if (value.nodeType === 3) {
+      retval += value.nodeValue;
+    }
+    if (value.nodeType !== 1 || (value as Element).tagName !== 'DEL') {
```
bakkot, line 152: Presumably this can be an `else if`, yes?

src/lint/rules/algorithm-line-endings.ts
```diff
@@ -29,11 +29,29 @@ export default function (report: Reporter, node: Element): Observer {
...
+      let first = node.contents[firstIndex];
+      let last = node.contents[lastIndex];
+      while (
+        first.name === 'tag' &&
+        last.name === 'tag' &&
+        ((first.contents === '<mark>' && last.contents === '</mark>') ||
+          (first.contents === '<ins>' && last.contents === '</ins>') ||
+          (first.contents === '<del>' && last.contents === '</del>'))
+      ) {
+        first = node.contents[++firstIndex];
```
bakkot, line 47: Nit: I am not a fan of usinig `++` and `--` in expression position.

test/lint.js
```diff
@@ -237,6 +237,9 @@ describe('linting whole program', function () {
           <emu-clause id="i7">
             <h1>Function ( _p1_, _p2_, &hellip; , _pn_, _body_ )</h1>
           </emu-clause>
+          <emu-clause id="foo">
+            <h1>Example ( _a_, <del>_b_</del><ins>_c_</ins> )</h1>
```
bakkot, line 241: Can you add a test showing that errors still work as well?

src/lint/collect-nodes.ts
ExE-Boss: That would change the behaviour of `nodeType === 3`, as that currently matches both `if` statements.

bakkot: It does match both, but text nodes can't have children, so there is no point in matching the second one.

ptomato:
Thanks for the review. I won't be able to revise this next week but will do it as soon as possible after that.

test/lint.js
ptomato: I added one checking that `Example ( _a_ [ , <del>_b_</del> ] )` gives an error — is that the sort of test you meant?

ptomato:
OK, this has been updated.

test/lint.js
ExE-Boss: You should probably also add a test that `Example ( _a_ <del>[ , _b_ ]</del> )` and `Example ( _a_ <ins>[ , _b_ ]</ins> )` work correctly.

ptomato: Good suggestion; by doing that I found out that leading and trailing spaces inside `<ins>` and `<del>` tags should be allowed, so I've added another commit changing that.

src/lint/collect-nodes.ts
```diff
@@ -140,3 +140,15 @@ export function collectNodes(
 
   return { success: true, mainGrammar, headers, sdos, earlyErrors, algorithms };
 }
+
+function textContentExcludingDeleted(node: Node): string {
+  let retval = '';
+  node.childNodes.forEach(value => {
+    if (value.nodeType === 3) {
+      retval += value.nodeValue;
+    } else if (value.nodeType !== 1 || (value as Element).tagName !== 'DEL') {
```
michaelficarra, line 149: I think you should only descend into [flow content](https://html.spec.whatwg.org/multipage/dom.html#flow-content) (excluding `<template>`?). Not sure the easiest way to do that.

bakkot: Ehh, I don't think that's worth worrying about in this PR.

Issue #192, opened 2020-04-26
Support referencing algorithm steps


bakkot:
Currently [emu-xref](https://tc39.es/ecmarkup/#emu-xref) allows linking to specific algorithms, but sometimes (especially in Annex B) we want to talk about specific steps of those algorithms. There's no way to do that automatically, which means we write out step numbers, and those inevitably get stale. It would be good if there were some way to give names to certain algorithm steps and then refer to those names and have that be translated to the appropriate step number.

ExE-Boss:
Duplicate of <https://github.com/tc39/ecmarkdown/issues/56>

ExE-Boss:
See also: <https://github.com/tc39/ecmarkdown/issues/56>.

PR #217, opened 2020-06-16
support for replacement algorithms and for xrefs to steps

bakkot:
This adds the ability to reference steps, both with `emu-xref`s and by marking algorithms as replacing steps (the latter so that they can have the appropriate numbering).

You can label steps using the `1. [label="foo"] Step.` syntax introduced in https://github.com/tc39/ecmarkdown/pull/74, and you can refer to them with `<emu-xref href="#step-foo"></emu-xref>` or `<emu-alg replaces-step="foo">`. The xref will automatically be populated with the appropriate step number (just the number itself, so you still need to write out "step", and possibly the name of the algorithm if it isn't clear). The labeled algorithm will automatically adopt the numbering of the labeled step, including for nested steps.

Now that we have support for replacement algorithms there is no longer a need to manually override the starting number for a list. I enforce that you do not, and that you do not write "step 1" or whatever, with the linter.

Fixes https://github.com/tc39/ecmarkup/issues/192.

Marked as draft until a version of ecmarkdown with support for the `[label=""]` syntax is published.

Corresponding ecmarkdown PR at https://github.com/tc39/ecmarkdown/pull/74. Corresponding ecma262 PR at https://github.com/tc39/ecma262/pull/2052.

Since it's hard to review the CSS in the PR, here's a screenshot of the new `algorithm-replacements` baseline:

Screen Shot 2020-06-15 at 9 34 16 PM

### DIFF
--- a/css/elements.css
+++ b/css/elements.css
@@ -78,16 +78,43 @@ emu-const {
 emu-val {
   font-weight: bold;
 }
-emu-alg ol, emu-alg ol ol ol ol {
-    list-style-type: decimal;
+
+/* depth 1 */
+emu-alg ol,
+/* depth 4 */
+emu-alg ol ol ol ol,
+emu-alg ol.nested-thrice,
+emu-alg ol.nested-twice ol,
+emu-alg ol.nested-once ol ol {
+  list-style-type: decimal;
 }
 
-emu-alg ol ol, emu-alg ol ol ol ol ol {
-    list-style-type: lower-alpha;
+/* depth 2 */
+emu-alg ol ol,
+emu-alg ol.nested-once,
+/* depth 5 */
+emu-alg ol ol ol ol ol,
+emu-alg ol.nested-four-times,
+emu-alg ol.nested-thrice ol,
+emu-alg ol.nested-twice ol ol,
+emu-alg ol.nested-once ol ol ol {
+  list-style-type: lower-alpha;
 }
 
-emu-alg ol ol ol, ol ol ol ol ol ol {
-    list-style-type: lower-roman;
+/* depth 3 */
+emu-alg ol ol ol,
+emu-alg ol.nested-twice,
+emu-alg ol.nested-once ol,
+/* depth 6 */
+emu-alg ol ol ol ol ol ol,
+emu-alg ol.nested-lots,
+emu-alg ol.nested-four-times ol,
+emu-alg ol.nested-thrice ol ol,
+emu-alg ol.nested-twice ol ol ol,
+emu-alg ol.nested-once ol ol ol ol,
+/* depth 7+ */
+emu-alg ol.nested-lots ol {
+  list-style-type: lower-roman;
 }
 
 emu-eqn {
--- a/src/Algorithm.ts
+++ b/src/Algorithm.ts
@@ -1,5 +1,7 @@
 import type { Context } from './Context';
+import type { StepBiblioEntry } from './Biblio';
 
+import { logWarning } from './utils';
 import Builder from './Builder';
 import * as emd from 'ecmarkdown';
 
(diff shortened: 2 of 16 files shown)


src/Algorithm.ts
```diff
@@ -9,22 +11,61 @@ export default class Algorithm extends Builder {
...
+        element: node,
+        target: replaces,
+      });
+      context.spec.replacementAlgorithmToContainedLabeledStepEntries.set(node, labeledStepEntries);
+    }
+
+    let labeledSteps = Array.from(node.querySelectorAll('li[id]'));
+    if (replaces && labeledSteps.length > 0 && node.firstElementChild!.children.length > 1) {
+      logWarning(
+        'You should not label a step in an replacement algorithm which has multiple top-level steps because the resulting step number could be ambiguous.'
```
michaelficarra, line 38:
```suggestion
        'You should not label a step in a replacement algorithm which has multiple top-level steps because the resulting step number could be ambiguous.'
```

css/elements.css
```diff
@@ -78,16 +78,35 @@ emu-const {
...
+  list-style-type: decimal;
+}
+
+emu-alg ol ol,
+emu-alg ol ol ol ol ol,
+emu-alg ol.nested-once,
+emu-alg ol.nested-thrice ol,
+emu-alg ol.nested-twice ol ol,
+emu-alg ol.nested-four-times,
+emu-alg ol.nested-once ol ol ol {
```
michaelficarra, line 96: Re-ordering this makes it easier for me to verify.

```suggestion
emu-alg ol ol,
emu-alg ol.nested-once,
emu-alg ol ol ol ol ol,
emu-alg ol.nested-four-times,
emu-alg ol.nested-thrice ol,
emu-alg ol.nested-twice ol ol,
emu-alg ol.nested-once ol ol ol {
```

css/elements.css
```diff
@@ -78,16 +78,35 @@ emu-const {
...
+
+emu-alg ol ol ol,
+emu-alg ol ol ol ol ol ol,
+emu-alg ol.nested-twice,
+emu-alg ol.nested-once ol,
+emu-alg ol.nested-four-times ol,
+emu-alg ol.nested-thrice ol ol,
+emu-alg ol.nested-twice ol ol ol,
+emu-alg ol.nested-once ol ol ol ol,
+emu-alg ol.nested-lots {
```
michaelficarra, line 108:
```suggestion
emu-alg ol ol ol,
emu-alg ol.nested-twice,
emu-alg ol.nested-once ol,
emu-alg ol ol ol ol ol ol,
emu-alg ol.nested-lots,
emu-alg ol.nested-four-times ol,
emu-alg ol.nested-thrice ol ol,
emu-alg ol.nested-twice ol ol ol,
emu-alg ol.nested-once ol ol ol ol,
emu-alg ol.nested-lots ol {
```

Did some re-ordering and also I think you missed `emu-alg ol.nested-lots ol`.

bakkot: > `emu-alg ol.nested-lots ol`

That shouldn't be necessary; it's implied by `ol.nested-lots`. No?

michaelficarra: According to my reading, the inner `<ol>` of `<emu-alg><ol class="nested-lots><ol> ...` would only be matched by `emu-alg ol ol`, which would give it `list-style-type: lower-alpha` not `list-style-type: lower-roman`.

src/Spec.ts
```diff
@@ -625,6 +633,70 @@ export default class Spec {
...
+        if (stepNumbers.length === 2) {
+          rootList.classList.add('nested-once');
+        } else if (stepNumbers.length === 3) {
+          rootList.classList.add('nested-twice');
+        } else if (stepNumbers.length === 4) {
+          rootList.classList.add('nested-thrice');
+        } else if (stepNumbers.length === 5) {
+          rootList.classList.add('nested-four-times');
+        } else {
+          // At the sixth level and deeper (so once we have nested five times) we just use `i`, so no further cases are necessary
```
michaelficarra, line 654:
```suggestion
          // At the sixth level and deeper (so once we have nested five times) we use a consistent line numbering style, so no further cases are necessary
```

src/Xref.ts
```diff
@@ -198,6 +203,33 @@ function buildFigureLink(
   }
 }
 
+let decimalBullet = Array.from({ length: 100 }).map((a, i) => '' + (i + 1));
+let alphaBullet = Array.from({ length: 26 }).map((a, i) =>
+  String.fromCharCode('a'.charCodeAt(0) + i)
+);
+// prettier-ignore
+let romanBullet = ['i', 'ii', 'iii', 'iv', 'v', 'vi', 'vii', 'viii', 'ix', 'x', 'xi', 'xii', 'xiii', 'xiv', 'xv', 'xvi', 'xvii', 'xviii', 'xix', 'xx', 'xxi', 'xxii', 'xxiii', 'xxxiv', 'xv'];
```
michaelficarra, line 211:
```suggestion
let romanBullet = ['i', 'ii', 'iii', 'iv', 'v', 'vi', 'vii', 'viii', 'ix', 'x', 'xi', 'xii', 'xiii', 'xiv', 'xv', 'xvi', 'xvii', 'xviii', 'xix', 'xx', 'xxi', 'xxii', 'xxiii', 'xxiv', 'xxv'];
```

?

css/elements.css
bakkot: Yup. Updated.

src/Xref.ts
```diff
@@ -198,6 +203,33 @@ function buildFigureLink(
   }
 }
 
+let decimalBullet = Array.from({ length: 100 }).map((a, i) => '' + (i + 1));
+let alphaBullet = Array.from({ length: 26 }).map((a, i) =>
+  String.fromCharCode('a'.charCodeAt(0) + i)
+);
+// prettier-ignore
+let romanBullet = ['i', 'ii', 'iii', 'iv', 'v', 'vi', 'vii', 'viii', 'ix', 'x', 'xi', 'xii', 'xiii', 'xiv', 'xv', 'xvi', 'xvii', 'xviii', 'xix', 'xx', 'xxi', 'xxii', 'xxiii', 'xxiv', 'xxv'];
+let bullets = [decimalBullet, alphaBullet, romanBullet, decimalBullet, alphaBullet, romanBullet];
```
michaelficarra, line 212: Why do we have to use stylised step numbers here? Regular Indic/Arabic numerals should do just fine, right? I hate all this complexity and I hate that it's coupled with the styles.

bakkot: Personally I find "step 6.a.vi" a lot nicer than "step 6.1.3". I'm not dead set against changing it, but I don't think it should be changed as part of this PR; feel free to bring it up on the editor call.

src/lint/collect-spelling-diagnostics.ts
```diff
@@ -51,6 +51,10 @@ let matchers = [
     pattern: /\r/gu,
     message: 'Only Unix-style (LF) linebreaks are allowed',
   },
+  {
+    pattern: /(?<=\b[Ss]teps? )\d/gu,
```
michaelficarra, line 55: How does this not trigger all over [Algorithm Conventions](https://tc39.es/ecma262/#sec-algorithm-conventions) where we describe how to read algorithm steps?

src/Algorithm.ts
```diff
@@ -9,22 +11,61 @@ export default class Algorithm extends Builder {
...
 
     // replace spaces after !/? with &nbsp; to prevent bad line breaking
-    const html = emd
-      .algorithm(node.innerHTML)
-      .replace(/((?:\s+|>)[!?])\s+(\w+\s*\()/g, '$1&nbsp;$2');
+    const html = rawHtml.replace(/((?:\s+|>)[!?])\s+(\w+\s*\()/g, '$1&nbsp;$2');
     node.innerHTML = html;
+
+    let labeledStepEntries: StepBiblioEntry[] = [];
+    let replaces = node.getAttribute('replaces-step');
```
michaelficarra, line 25: Should we remove this attribute from the output?

src/Xref.ts
michaelficarra: Oh I was thinking these were for internal references for some reason. I see now that it's for the text in the reference link. Yeah I guess this is fine for now. I wish we could somehow coordinate this logic with the CSS styling.

src/Algorithm.ts
bakkot: Ecmarkup doesn't really clean up after itself, as a rule - all the `emu-whatever` tags are in the output, etc.

michaelficarra: Fine. We should reconsider that.

bakkot: Opened https://github.com/tc39/ecmarkup/issues/218 to track.

PR #219, opened 2020-06-18
use id rather than label

bakkot:
Followup to https://github.com/tc39/ecmarkdown/pull/75.

Do we want to introduce an editorial convention that all step labels must be prefixed with `step-`, now that tooling doesn't do that for us? I can add a lint rule to enforce that, if so.

Contains version bump commit, so please **rebase**, not squash.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -734,9 +734,9 @@
       }
     },
     "ecmarkdown": {
-      "version": "5.1.0",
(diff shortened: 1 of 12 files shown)


bakkot:
OK, added a lint rule (and bumped ecmarkdown to fix an issue with locations). If this looks good I'll squash it down to just the two commits and merge.

PR #220, opened 2020-06-20
Fix <emu‑biblio> documentation

ExE-Boss:
This also fixes the auto‑link for `Get` to point to <https://tc39.es/ecma262/#sec-get-o-p> instead of <https://tc39.github.io/ecma262/#sec-get-o-p>

### DIFF
--- a/spec/biblio.json
+++ b/spec/biblio.json
@@ -1,8 +1,9 @@
 {
-  "https://tc39.github.io/ecma262/": [
+  "https://tc39.es/ecma262/": [
     {
       "type": "op",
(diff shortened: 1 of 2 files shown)


spec/biblio.json
```diff
@@ -1,8 +1,9 @@
 {
-  "https://tc39.github.io/ecma262/": [
+  "https://tc39.es/ecma262/": [
     {
       "type": "op",
-      "aoid": "ReturnIfAbrubt"
```
bakkot, line 5: ... How did this even happen

PR #223, opened 2020-06-27
[meta] add "engines"

### LINKED ISSUE #112: Node v4 is not actually supported
The problem is a phrase in lib/cli.js, `for (const [file, watcher] of watching) ...` which is probably too-new JS for Node v4.  Node v7.4.0 works for me, but it would probably be better if the TypeScript compiler could translate the phrase to Node v4 compatible JS.

ljharb:
Ran `npx ls-engines` to calculate the ranges.

Closes #112.

(since "engines" is purely advisory, this is semver-neutral; for yarn users, however, it will fail the install on node versions where it was already broken)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ A web component-based source format for ECMAScript and related specifications. S
 
 ### Installing & Using
 
-Requires Node.js 4.0 or above.
+Requires node 8 or above.
 
 ```
 npm install ecmarkup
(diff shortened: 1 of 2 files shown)


bakkot:
#112 refers to the fact that the readme says "Requires Node.js 4.0 or above.", so this does not actually fix that issue.

ljharb:
Fair; for simplicity i updated the readme to say "node 8".

Issue #226, opened 2020-07-10
Better errors


bakkot:
For example, having multiple linebreaks in a `<p>` should not abort the build with "Error: expecting EOF, got parabreak".

At the very least we need to start wrapping those and providing location information, because the errors at the moment are totally useless. But also we should clean up the output to suggest what's actually wrong. And in the above case it shouldn't be a hard error at all, just a lint warning.

bakkot:
Thinking about this some: I'd like to use the same format for other errors as we do for linting errors. But the eslint error formatters don't support streaming: specifically, they always add a "1 error found." summary at the end, which would be weird to see multiple times. I'm not entirely sure how to handle it. Here's some possibilities, in roughly descending order of how much I like them

- drop support for arbitrary formatters and reimplement the codeframe formatter
- drop support for arbitrary formatters and manually modify the output from codeframe to not have the summary
- support arbitrary formatters but try to tweak their output anyway
- support arbitrary formatters and just not worry about the confusing output
- not report any errors until the end of the build, so we can report them all at once

Maybe there's some other alternative?

PR #229, opened 2020-07-22
Better errors

bakkot:
This is on top of https://github.com/tc39/ecmarkup/pull/228 and is marked as a draft until that lands. https://github.com/tc39/ecmarkup/issues/226 remains open until a subsequent PR converts ecmarkdown failures into warnings.

This provides filename/line/column information for everything which is currently a warning, and unifies linting errors with other errors. The CLI wrapper for the library prints warnings out the same way it does currently (in red, as they are encountered), except with file/line/column. If you pass --verbose it also collects them at the end in sorted order using the provided eslint formatter.

This is a breaking change, at least if you consider ecmarkup to be a library and not just a CLI tool.

Screenshot (the 'FAIL: 1' is my prompt printing the exit status, not part of emarkup's output):

Screen Shot 2020-07-23 at 11 19 45 AM

<details>
<summary>screenshot without --strict</summary>
Screen Shot 2020-07-23 at 11 27 08 AM

</details>


src/Spec.ts
```diff
@@ -408,7 +493,12 @@ export default class Spec {
     try {
       data = yaml.safeLoad(block.textContent!);
     } catch (e) {
```
michaelficarra, line 495: Does this error contain position info we can preserve?

test/errors.js
```diff
@@ -0,0 +1,304 @@
...
+      `,
+      {
+        ruleId: 'invalid-replacement',
+        nodeType: 'emu-alg',
+        message: 'expected algorithm to replace a step, not a clause',
+      }
+    );
+  });
+
+  it('cycle in replacement algorithms', async () => {
```
michaelficarra, line 70: Can you add a bigger (non-self-referential) cycle test?

src/Spec.ts
bakkot: Technically yes, though it's not particularly good information. Updated.

test/errors.js
bakkot: Done.

PR #233, opened 2020-07-24
Update operation name linting to allow slashes

### LINKED ISSUE #232: Linter rejects names like "ForIn/OfHeadEvaluation" unless preceded by "{Runtime,Static} Semantics"
cf. https://github.com/tc39/ecma262/pull/2110#issuecomment-663531308
```
error: expected operation to have a name like 'Example', 'Runtime Semantics: Foo', 'Example.prop', etc, but found "ForIn/OfHeadEvaluation " (header-format) at spec.html:18104:13:
  18102 | 
  18103 |       <emu-clause id="sec-runtime-semantics-forinofheadevaluation" oldids="sec-runtime-semantics-forin-div-ofheadevaluation-tdznames-expr-iterationkind" aoid="ForIn/OfHeadEvaluation">
> 18104 |         <h1>ForIn/OfHeadEvaluation ( _uninitializedBoundNames_, _expr_, _iterationKind_ )</h1>
        |             ^
  18105 |         <p>The abstract operation ForIn/OfHeadEvaluation takes arguments _uninitializedBoundNames_, _expr_, and _iterationKind_ (either ~enumerate~, ~iterate~, or ~async-iterate~). It performs the following steps when called:</p>
  18106 |         <emu-alg>
  18107 |           1. Let _oldEnv_ be the running execution context's LexicalEnvironment.

```

gibson042:
Fixes #232

### DIFF
--- a/test/lint.js
+++ b/test/lint.js
@@ -132,17 +132,14 @@ describe('linting whole program', function () {
     it('name format', async function () {
       await assertLint(
         positioned`
-          <emu-clause id="a">
-            <h1>unusual/valid ( )</h1>
-          </emu-clause>
-          <emu-clause id="b">
-            <h1>${M}foo: ( )</h1>
+          <emu-clause id="foo">
+            <h1>${M}something: ( )</h1>
         `,
         {
           ruleId: 'header-format',
           nodeType: 'h1',
           message:
-            "expected operation to have a name like 'Example', 'Runtime Semantics: Foo', 'Example.prop', etc, but found \"foo: \"",
+            "expected operation to have a name like 'Example', 'Runtime Semantics: Foo', 'Example.prop', etc, but found \"something: \"",
         }
       );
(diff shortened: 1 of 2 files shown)


test/lint.js
```diff
@@ -132,14 +132,17 @@ describe('linting whole program', function () {
     it('name format', async function () {
       await assertLint(
         positioned`
-          <emu-clause id="foo">
-            <h1>${M}something: ( )</h1>
+          <emu-clause id="a">
```
bakkot, line 135: Rather than modifying this test, this should modify the [legal names](https://github.com/tc39/ecmarkup/blob/da698c3b4418e33ed3bd1afdbcac8fd87a0f0bed/test/lint.js#L188) test below.

gibson042: Thanks, I missed that one. Updated.

PR #234, opened 2020-07-26
Wrap ecmarkdown parse failures

bakkot:
This wraps ecmarkdown parse failures with the same error-reporting mechanism used elsewhere (see https://github.com/tc39/ecmarkup/pull/229).

Previously when a node failed to parse with Ecmarkdown, Ecmarkup threw an exception. Now it emits a warning, and puts the raw source for that node (wrapped in `#### ECMARKDOWN PARSE FAILED ###<pre>` ) in the output.

This means that a.) you get useful location information (including a little snippet of context if you pass `--verbose`) instead of just being told something somewhere broke and b.) you still get a build artifact you can review (except for the broken part), unless you pass `--strict`.

Fixes https://github.com/tc39/ecmarkup/issues/226.

### DIFF
--- a/src/Algorithm.ts
+++ b/src/Algorithm.ts
@@ -3,6 +3,7 @@ import type { Node as EcmarkdownNode, OrderedListItemNode } from 'ecmarkdown';
 import type { StepBiblioEntry } from './Biblio';
 
 import Builder from './Builder';
+import { warnEmdFailure } from './utils';
 import * as emd from 'ecmarkdown';
 
 function findLabeledSteps(root: EcmarkdownNode) {
@@ -25,11 +26,20 @@ export default class Algorithm extends Builder {
 
     let innerHTML = node.innerHTML; // TODO use original slice, forward this from linter
 
-    // prettier-ignore
-    const emdTree =
-      'ecmarkdownTree' in node
-        ? (node as any).ecmarkdownTree
-        : emd.parseAlgorithm(innerHTML);
+    let emdTree;
+    if ('ecmarkdownTree' in node) {
+      emdTree = (node as any).ecmarkdownTree;
+    } else {
+      try {
+        emdTree = emd.parseAlgorithm(innerHTML);
+      } catch (e) {
+        warnEmdFailure(spec.warn, node, e);
+      }
+    }
+    if (emdTree == null) {
+      node.innerHTML = `#### ECMARKDOWN PARSE FAILED ###<pre>${innerHTML}</pre>`
+      return;
+    }
 
     const rawHtml = emd.emit(emdTree);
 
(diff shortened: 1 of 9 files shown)


src/Algorithm.ts
```diff
@@ -25,11 +26,20 @@ export default class Algorithm extends Builder {
...
+      emdTree = (node as any).ecmarkdownTree;
+    } else {
+      try {
+        emdTree = emd.parseAlgorithm(innerHTML);
+      } catch (e) {
+        warnEmdFailure(spec.warn, node, e);
+      }
+    }
+    if (emdTree == null) {
+      node.innerHTML = `#### ECMARKDOWN PARSE FAILED ###<pre>${innerHTML}</pre>`;
```
michaelficarra, line 40:
```suggestion
      node.innerHTML = `### ECMARKDOWN PARSE FAILED ###<pre>${innerHTML}</pre>`;
```

michaelficarra: There's a couple cases of these. You might want to extract this to a constant and share it. Also, for the record, I believe the proper way to do this would be to add this content using CSS `:before { content: ... }` since it's entirely presentational.

src/Spec.ts
```diff
@@ -116,46 +116,79 @@ export type Warning =
...
       column = undefined;
+      nodeType = 'html';
     } else if (e.type === 'raw') {
       ({ line, column } = e);
+      nodeType = 'html';
     } else {
-      let nodeLoc = utils.getLocation(dom, e.node);
       if (e.type === 'node') {
-        ({ line, col: column } = nodeLoc.startTag);
+        if (e.node.nodeType === 3) {
```
michaelficarra, line 131:
```suggestion
        if (e.node.nodeType === Node.TEXT_NODE) {
```

src/Spec.ts
```diff
@@ -116,46 +116,79 @@ export type Warning =
...
-          tagSrc,
-          nodeLoc.startTag.endOffset - nodeLoc.startOffset
-        );
-        line = nodeLoc.startTag.line + tagEnd.line + nodeRelativeLine - 2;
-        if (nodeRelativeLine === 1) {
-          if (tagEnd.line === 1) {
-            column = nodeLoc.startTag.col + tagEnd.column + nodeRelativeColumn - 2;
-          } else {
-            column = tagEnd.column + nodeRelativeColumn - 1;
+        if (e.node.nodeType === 3) {
```
michaelficarra, line 152:
```suggestion
        if (e.node.nodeType === Node.TEXT_NODE) {
```

src/Spec.ts
bakkot: That's not a global in Node, but I'll add a comment.

Issue #213, opened 2020-06-15
Linter should handle <emu-import>


ptomato:
It seems that if you have an `<emu-alg>` inside a file that is included with `<emu-import>`, the linter will choke:
**bug.html**
```html
<!doctype html>
<meta charset="utf8">

<pre class=metadata>
title: Bug proposal
stage: 1
contributors: E. S. Expert
</pre>

<emu-intro id=sec-intro>
  <h1>Introduction</h1>
</emu-intro>

<emu-import href="bug-sub.html"></emu-import>
```
**bug-sub.html**
```html
<!doctype html>
<meta charset="utf8">

<emu-clause id="sec-bug1">
  <h1>Bug</h1>
  <emu-alg>
    1. Let this be a bug.
  </emu-alg>
</emu-clause>
```
If you replace the `<emu-import>` with the `<emu-clause>` from bug-sub.html, then the linter succeeds and the ecmarkup file is rendered. (I found this while trying to lint the tc39/proposal-temporal spec, which is split up into several files.)

Stack trace:
```
Error: could not find location
    at Object.getLocation (/.../node_modules/ecmarkup/lib/lint/utils.js:48:15)
    at Object.collectAlgorithmDiagnostics (/.../node_modules/ecmarkup/lib/lint/collect-algorithm-diagnostics.js:38:32)
    at Object.lint (/.../node_modules/ecmarkup/lib/lint/lint.js:24:59)
    at Spec.<anonymous> (/.../node_modules/ecmarkup/lib/Spec.js:163:24)
    at Generator.next (<anonymous>)
    at fulfilled (/.../node_modules/ecmarkup/lib/Spec.js:5:58)
```

PR #235, opened 2020-08-05
Track location information for emu-import'd nodes

bakkot:
Fixes https://github.com/tc39/ecmarkup/issues/213.

`emu-import` works by parsing the referenced file and sticking its contents in the DOM. Errors in the resulting nodes need to be mapped back to the correct location in the imported file; they can't just be looked up in the origin jsdom object.

Rather than adding a bunch of new tests I just modified the existing test helper to also test the given error in an `emu-import`.

One oddity is that I modified `Import.ts` to insert nodes in the original DOM using `adoptNode` rather than `cloneNode` (so that identity is preserved for jsdom's location-querying method) and this change for some reason results in there not being a trailing linebreak before the closing `</emu-import>` when rendering the HTML. No idea what's up with that, but it shouldn't cause any issues. That's why the baselines changed.

### DIFF
--- a/src/lint/collect-algorithm-diagnostics.ts
+++ b/src/lint/collect-algorithm-diagnostics.ts
@@ -34,12 +34,12 @@ function composeObservers(...observers: Observer[]): Observer {
 export function collectAlgorithmDiagnostics(
   report: (e: Warning) => void,
   spec: Spec,
-  sourceText: string,
+  mainSource: string,
   algorithms: { element: Element; tree?: EcmarkdownNode }[]
 ) {
   for (let algorithm of algorithms) {
     let element = algorithm.element;
-    let { source, ...location } = spec.locate(element);
+    let { source: importSource, ...location } = spec.locate(element);
 
     if (location.endTag == null) {
       report({
@@ -63,7 +63,7 @@ export function collectAlgorithmDiagnostics(
       });
     };
 
-    let algorithmSource = (source ?? sourceText).slice(
+    let algorithmSource = (importSource ?? mainSource).slice(
       location.startTag.endOffset,
       location.endTag.startOffset
     );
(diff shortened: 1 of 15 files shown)


src/lint/collect-algorithm-diagnostics.ts
```diff
@@ -33,13 +33,13 @@ function composeObservers(...observers: Observer[]): Observer {
...
   report: (e: Warning) => void,
-  dom: any,
+  spec: Spec,
   sourceText: string,
   algorithms: { element: Element; tree?: EcmarkdownNode }[]
 ) {
   for (let algorithm of algorithms) {
     let element = algorithm.element;
-    let location = getLocation(dom, element);
+    let { source, ...location } = spec.locate(element);
```
michaelficarra, line 42: Having both `source` and `sourceText` in scope at the same time is a bit confusing. Consider renaming this?

src/utils.ts
```diff
@@ -155,6 +155,9 @@ export function offsetToLineAndColumn(string: string, offset: number) {
   let line = 0;
   let seen = 0;
   while (true) {
+    if (line >= lines.length) {
+      throw new Error(`offset ${offset} exceeded string ${JSON.stringify(string)}`);
```
michaelficarra, line 159: You sure you want to put the whole string here? That could be really long. Why not just put the maximum offset or refer to the lines or something?

bakkot: This is basically an assertion failure. The point of the error message is to have the whole string so I can see what I did wrong. I don't expect it to get surfaced to users.

src/lint/collect-algorithm-diagnostics.ts
bakkot: Renamed to `importSource` and `mainSource` respectively.

PR #239, opened 2020-08-18
Linting: misc improvements

bakkot:
Three new lint rules and a couple of tweaks to existing ones.

d746f13 pertains to https://github.com/tc39/ecma262/pull/2133. 814c4c9 pertains to https://github.com/tc39/ecma262/pull/2125.

It's probably easiest to review the commits individually, and I would rebase rather than squash this PR.

### DIFF
--- a/src/lint/collect-header-diagnostics.ts
+++ b/src/lint/collect-header-diagnostics.ts
@@ -46,7 +46,8 @@ export function collectHeaderDiagnostics(
       // Object.fromEntries
       // ForIn/OfHeadEvaluation
       // Array.prototype [ @@iterator ]
-      /^[A-Za-z][A-Za-z0-9/]*(\.[A-Za-z][A-Za-z0-9/]*)*( \[ @@[a-z][a-zA-Z]+ \])?\s*$/,
+      // Object.prototype.__defineGetter__
+      /^[A-Za-z][A-Za-z0-9/]*(\.[A-Za-z][A-Za-z0-9]*)*(\.__[a-z][A-Za-z0-9]*__| \[ @@[a-z][a-zA-Z]+ \])?\s*$/,
 
       // %ForInIteratorPrototype%.next
       // %TypedArray%.prototype [ @@iterator ]
--- a/src/lint/collect-spelling-diagnostics.ts
+++ b/src/lint/collect-spelling-diagnostics.ts
@@ -65,6 +65,14 @@ let matchers = [
     pattern: /(?<=\S)  +(?! |<\/(td|th|dd|dt)>)/gu,
     message: 'multiple consecutive spaces are not allowed',
   },
+  {
+    pattern: /(?<=<[a-z]+( [a-z]+(="[^"\n]+")?)*>)(?<!(td|th|dd|dt)>) /gu,
+    message: 'tags should not contain leading whitespace',
+  },
+  {
+    pattern: /(?<=[^ \n]) +<\/(?!td|th|dd|dt)/gu,
+    message: 'tags should not contain trailing whitespace',
+  },
 ];
 
 export function collectSpellingDiagnostics(
--- a/src/lint/collect-spelling-diagnostics.ts
+++ b/src/lint/collect-spelling-diagnostics.ts
@@ -56,6 +56,11 @@ let matchers = [
     pattern: /(?<=\b[Ss]teps? )\d/gu,
     message: 'prefer using labeled steps and <emu-xref> tags over hardcoding step numbers',
   },
+  {
+    pattern: /(?<=\b[Cc]lauses? )\d/gu,
+    message:
+      'clauses should be referenced using <emu-xref> tags rather than hardcoding clause numbers',
+  },
 ];
 
 export function collectSpellingDiagnostics(
--- a/src/lint/collect-spelling-diagnostics.ts
+++ b/src/lint/collect-spelling-diagnostics.ts
@@ -61,6 +61,10 @@ let matchers = [
     message:
       'clauses should be referenced using <emu-xref> tags rather than hardcoding clause numbers',
   },
+  {
+    pattern: /(?<=\S)  +(?! |<\/(td|th|dd|dt)>)/gu,
+    message: 'multiple consecutive spaces are not allowed',
+  },
 ];
 
 export function collectSpellingDiagnostics(
--- a/src/lint/collect-spelling-diagnostics.ts
+++ b/src/lint/collect-spelling-diagnostics.ts
@@ -9,19 +9,19 @@ let ruleId = 'spelling';
 let matchers = [
   {
     pattern: /\*this\* object/giu,
-    message: 'Prefer "*this* value"',
+    message: 'prefer "*this* value"',
   },
   {
     pattern: /1's complement/giu,
-    message: 'Prefer "one\'s complement"',
+    message: 'prefer "one\'s complement"',
   },
   {
     pattern: /2's complement/giu,
-    message: 'Prefer "two\'s complement"',
+    message: 'prefer "two\'s complement"',
   },
   {
     pattern: /\*0\*(?!<sub>ℤ<\/sub>)/gu,
-    message: 'The Number value 0 should be written "*+0*", to unambiguously exclude "*-0*"',
+    message: 'the Number value 0 should be written "*+0*", to unambiguously exclude "*-0*"',
   },
   {
     pattern: /behavior/giu,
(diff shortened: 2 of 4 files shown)


src/lint/collect-spelling-diagnostics.ts
```diff
@@ -56,6 +56,10 @@ let matchers = [
     pattern: /(?<=\b[Ss]teps? )\d/gu,
     message: 'prefer using labeled steps and <emu-xref> tags over hardcoding step numbers',
   },
+  {
+    pattern: /(?<=\S)  +(?! |<\/t(d|h)>)/gu,
+    message: 'multiple consecutive spaces are not allowed',
```
ljharb, line 61:
```suggestion
    message: 'multiple consecutive non-indentation spaces are not allowed',
```

src/lint/collect-spelling-diagnostics.ts
```diff
@@ -59,7 +59,12 @@ let matchers = [
   {
     pattern: /(?<=\S)  +(?! |<\/t(d|h)>)/gu,
     message: 'multiple consecutive spaces are not allowed',
-  }
+  },
+  {
+    pattern: /[Cc]lauses? \d/gu,
+    message:
+      'clauses should be referenced using <emu-xref> tags rather than hardcoding clause numbers',
```
ljharb, line 66: should we also cover steps, not just clauses?

src/lint/collect-spelling-diagnostics.ts
```diff
@@ -65,6 +65,10 @@ let matchers = [
     message:
       'clauses should be referenced using <emu-xref> tags rather than hardcoding clause numbers',
   },
+  {
+    pattern: /(?<=[^ \n]) +<\/p>/gu,
+    message: '<p> tags should not contain trailing whitespace',
```
ljharb, line 70: should any tags contain trailing whitespace?

src/lint/collect-spelling-diagnostics.ts
bakkot: That's more precise, but I don't think the extra verbiage is necessary, since this will only warn on non-indentation spaces.

src/lint/collect-spelling-diagnostics.ts
bakkot: [Already done](https://github.com/tc39/ecmarkup/blob/f3ade26235472c57e7ab5972dcdf0aadb31fea34/src/lint/collect-spelling-diagnostics.ts#L56).

src/lint/collect-spelling-diagnostics.ts
bakkot: `<td>` and `<th>` can have trailing whitespace for aligning cells in the source. Can't think of any others off the top of my head.

ljharb: can have, but should they? html compresses all consecutive whitespace down to a single space, so any spaces over one would be redundant, no?

bakkot: It's for alignment _in the source_. Specifically, we have some tables in 262 which are written like

```html
<figure>
  <table class="lightweight-table">
    <tbody>
      <tr><th> _assignmentOpText_ </th><th> _opText_       </th></tr>
      <tr><td> `**=`              </td><td> `**`           </td></tr>
      <tr><td> `*=`               </td><td> `*`            </td></tr>
      <tr><td> `/=`               </td><td> `/`            </td></tr>
      <tr><td> `%=`               </td><td> `%`            </td></tr>
      <tr><td> `+=`               </td><td> `+`            </td></tr>
      <tr><td> `-=`               </td><td> `-`            </td></tr>
      <tr><td> `&lt;&lt;=`        </td><td> `&lt;&lt;`     </td></tr>
      <tr><td> `&gt;&gt;=`        </td><td> `&gt;&gt;`     </td></tr>
      <tr><td> `&gt;&gt;&gt;=`    </td><td> `&gt;&gt;&gt;` </td></tr>
      <tr><td> `&amp;=`           </td><td> `&amp;`        </td></tr>
      <tr><td> `^=`               </td><td> `^`            </td></tr>
      <tr><td> `|=`               </td><td> `|`            </td></tr>
    </tbody>
  </table>
</figure>
```
which I don't want the linter to forbid.

ljharb: ah, ok. that seems pretty gross to me in the source, but i'll ack that you're not trying to block it now :-)

test/lint-spelling.js
```diff
@@ -206,7 +219,11 @@ windows:${M}\r
       <emu-alg>
         1. [id="step-label"] Foo.
       </emu-alg>
-      Something about step <emu-xref href="#step-label"></emu-xref>.
+      <p>Something about step <emu-xref href="#step-label"></emu-xref>.</p>
+
+      <table><tr><td>Spaces for aligning table cells are acceptable.</td>
+      <td>For example, it is nice for these to line up.             </td>
```
michaelficarra, line 252: I don't really care to add this exception. If we did this, I'd think we'd also want to allow `</li>`, `</dt>`, `<dd>`, etc. So probably either allow any whitespace until `<` or disallow all internal whitespace. Is this exception necessary for ecma262 to pass currently?

src/lint/collect-spelling-diagnostics.ts
```diff
@@ -59,7 +59,12 @@ let matchers = [
   {
     pattern: /(?<=\S)  +(?! |<\/t(d|h)>)/gu,
     message: 'multiple consecutive spaces are not allowed',
-  }
+  },
+  {
+    pattern: /[Cc]lauses? \d/gu,
```
michaelficarra, line 64: We could use a lookbehind here so that the error points to the number, which is the real issue.
```suggestion
    pattern: /(?<=[Cc]lauses? )\d+/gu,
```

src/lint/collect-header-diagnostics.ts
```diff
@@ -46,7 +46,7 @@ export function collectHeaderDiagnostics(
       // Object.fromEntries
       // ForIn/OfHeadEvaluation
       // Array.prototype [ @@iterator ]
```
michaelficarra, line 48: Update this comment.

src/lint/collect-header-diagnostics.ts
```diff
@@ -46,7 +46,7 @@ export function collectHeaderDiagnostics(
...
-      /^[A-Za-z][A-Za-z0-9/]*(\.[A-Za-z][A-Za-z0-9/]*)*( \[ @@[a-z][a-zA-Z]+ \])?\s*$/,
+      /^[A-Za-z][A-Za-z0-9/]*(\.[A-Za-z][A-Za-z0-9]*)*((\.__[a-z][A-Za-z0-9]*__)?|( \[ @@[a-z][a-zA-Z]+ \])?)\s*$/,
```
michaelficarra, line 49:
```suggestion
      /^[A-Za-z][A-Za-z0-9/]*(\.[A-Za-z][A-Za-z0-9]*)*(\.__[a-z][A-Za-z0-9]*__| \[ @@[a-z][a-zA-Z]+ \])?\s*$/,
```

Don't put two optional things in a disjunction.

src/lint/collect-spelling-diagnostics.ts
michaelficarra, line 70: Nor leading whitespace, right? And why is `<p>` special here?

test/lint-spelling.js
bakkot: > Is this exception necessary for ecma262 to pass currently?

Yes, see [discussion above](https://github.com/tc39/ecmarkup/pull/239#discussion_r471938869).

src/lint/collect-spelling-diagnostics.ts
bakkot: I'll add leading whitespace as well.

`<p>` is special because it's the tag where this is the most likely to come up. Some tags like `<td>` intentionally contain trailing whitespace, so I don't want to do this for everything, and I don't want to come up with a full list of exactly which tags should and should not allow whitespace. If you have an uncontroversial version of such a list on hand, I'm happy to adopt it.

test/lint-spelling.js
michaelficarra: Then I think we should add some more of those other reasonable tags to the list or just forbid it entirely.

bakkot: I am happy to add other tags as well where it makes sense to do so, though I disagree that any of `</li>`,  `</dt>,` or `</dd>` would be good candidates: I would not expect to see multiple such tags on a single line, and so would not expect to see internal spaces used for alignment. Do you have other tags you'd like to add?

src/lint/collect-spelling-diagnostics.ts
bakkot: I suppose I'd be fine with matching the list I used for the multiple consecutive spaces rule, i.e. forbidding leading / trailing spaces in everything except `<td>` and `<th>`, if you'd prefer that.

src/lint/collect-header-diagnostics.ts
bakkot: Pulled the `?` to outside the disjunction.

src/lint/collect-header-diagnostics.ts
bakkot: Done.

src/lint/collect-spelling-diagnostics.ts
bakkot: Done.

test/lint-spelling.js
```diff
@@ -249,7 +262,8 @@ windows:${M}\r
       <p>Something about step <emu-xref href="#step-label"></emu-xref>.</p>
 
       <table><tr><td>Spaces for aligning table cells are acceptable.</td>
-      <td>For example, it is nice for these to line up.             </td>
+      <td> For example, it is nice for these to line up. </td>
+      <td> Also to pad them a biit, in some cases.       </td>
```
michaelficarra:
```suggestion
      <td> Also to pad them a bit, in some cases.        </td>
```

bakkot: Already done.

src/lint/collect-header-diagnostics.ts
```diff
@@ -46,7 +46,8 @@ export function collectHeaderDiagnostics(
       // Object.fromEntries
       // ForIn/OfHeadEvaluation
       // Array.prototype [ @@iterator ]
-      /^[A-Za-z][A-Za-z0-9/]*(\.[A-Za-z][A-Za-z0-9/]*)*( \[ @@[a-z][a-zA-Z]+ \])?\s*$/,
+      // Object.prototype.__defineGetter__
+      /^[A-Za-z][A-Za-z0-9/]*(\.[A-Za-z][A-Za-z0-9]*)*((\.__[a-z][A-Za-z0-9]*__)|( \[ @@[a-z][a-zA-Z]+ \]))?\s*$/,
```
michaelficarra, line 50:
```suggestion
      /^[A-Za-z][A-Za-z0-9/]*(\.[A-Za-z][A-Za-z0-9]*)*(\.__[a-z][A-Za-z0-9]*__| \[ @@[a-z][a-zA-Z]+ \])?\s*$/,
```

bakkot: Done.

PR #242, opened 2020-08-23
deploy from primary branch rather than pages branch

bakkot:
The repo includes a script to update gh-pages on every CI run, but a.) that's not the right time to update and b.) it's been broken for years. Instead, just serve the documentation from the main branch. I've added a script to .travis.yml to ensure that any changes to the spec are committed so it can't get stale.


.travis.yml
```diff
@@ -1,16 +1,10 @@
-before_install:
-  - nvm install-latest-npm
```
ljharb, line 2: this should be kept to ensure all relevant npm bugfixes are available.

.travis.yml
```diff
@@ -1,16 +1,10 @@
...
 script:
   - npm run build
+  - npm run build-spec
+  - if [ ! -z "$(git status --porcelain)" ]; then echo "repo is dirty; you probably need to 'npm run build-spec' and commit the result"; exit 1; fi
   - npm test
   - npm run lint
   - npm run test-published-files
```
ljharb, line 7: this is a lot of things to run in serial, they should probably be split into separate jobs soon (doesn't have to be in this PR)

.travis.yml
bakkot: Is travis not going to use the latest npm? That's unfortunate.

ljharb: no CI systems ever do by default; the npm version is only the one that comes with the CI job's node version, unless you manually update it; however, `npm install -g npm` isn't reliable, which is why `nvm install-latest-npm` exists.

PR #244, opened 2020-08-26
always define color for background & font together

imme-emosol:
The main thing I would like to have fixed is that I can easily read the text on the page as soon as it loads.
Currently the dark font color does not read well on the (dark) background-color I set in my browser.
A practice to make sure that elements will be colored as intended is to always declare background-color & color together.
Whether the transparent, initial & inherit are chosen well is something I do not know, I just did a search & replace for background or text color declared without the other.


michaelficarra:
Can't this be accomplished by just setting the document's background colour?

```css
html {
    background-color: #fff;
}
```

imme-emosol:
Better readability might be achieved by doing just that.

The idea about declaring color & background-color as an inseparable set, is that it clearly communicates to the reader of the style declarations that those colors are intended to be like that. While omitting either of those two properties leaves room for interpretation, ambiguity.

Another thing that seems to be used more frequent, is to declare variables for colors on the `:root` element and then use `var(--color-var-name);` for color & background-color. Which is then often combined with `prefers-color-scheme`. Or also another mechanism to toggle between color schemes (for instance a "normal" vs. high-contrast scheme of colors).

The main reason for making this merge request is to trigger some sort of re-iterating over current choices regarding the color scheme.

michaelficarra:
@imme-emosol I opened #545 to address this in the naïve way for now. Maybe in the future we can have multiple colour schemes and support `prefers-color-scheme`.

Issue #247, opened 2020-09-06
'ins' and 'del' do not render correctly in a simplified 'oneof' production


rbuckton:
In ecmarkup, the following syntax is allowed for a "one of" list:

**input**
```html
<emu-production name="A" type="lexical" oneof>
    <emu-rhs>a b c</emu-rhs>
</emu-production>
```

Which is parsed and emitted as the following:

**output**
```html
<emu-production name="A" type="lexical" oneof="" id="prod-A">
  <emu-nt><a href="#prod-A">A</a></emu-nt>
  <emu-geq>::</emu-geq>
  <emu-oneof>one of</emu-oneof>
  <emu-rhs>
    <emu-t>a</emu-t>
    <emu-t>b</emu-t>
    <emu-t>c</emu-t>
  </emu-rhs>
</emu-production>
```

However, if one of the terminals in the list is wrapped in `<ins>` or `<del>`, the resulting emit is incorrect:

**input**
```html
<emu-production name="A" type="lexical" oneof>
    <emu-rhs>a <ins>b</ins> c</emu-rhs>
</emu-production>
```

**output**
```html
<emu-production name="A" type="lexical" oneof="" id="prod-A">
  <emu-nt><a href="#prod-A">A</a></emu-nt>
  <emu-geq>::</emu-geq>
  <emu-oneof>one of</emu-oneof>
  <emu-rhs>
    <emu-t>a</emu-t>
    <ins>b</ins>
    <emu-t>c</emu-t>
  </emu-rhs>
</emu-production>
```

As you can see in the output, the terminal `b` is not enclosed in an `<emu-t>`.

Related issue: https://github.com/rbuckton/grammarkdown/issues/32

bakkot:
Opened https://github.com/tc39/ecmarkup/pull/249.

Issue #248, opened 2020-09-06
Grammar nodes copy-paste without whitespace


bakkot:
Because we use CSS margins to pad grammar nodes and do not include any actual whitespace, copy/pasting productions gives you all the text smooshed together without spaces. We should add some whitespace.

bakkot:
Ref: https://github.com/rbuckton/grammarkdown/issues/27

michaelficarra:
It also makes searching for grammar hard because you have to search without the whitespace. Most readers probably won't know this.

PR #249, opened 2020-09-06
support ins/del in emu-production tags

bakkot:
Fixes https://github.com/tc39/ecmarkup/issues/247 (cc @rbuckton).

Now

```html
<emu-production name="A" type="lexical" oneof>
  <emu-rhs><del>a</del> b c <mark>d</mark> <ins>e f</ins></emu-rhs>
</emu-production>
```
is emitted as (modulo whitespace)

```html
<emu-production name="A" type="lexical" oneof="" id="prod-A">
  <emu-nt><a href="#prod-A">A</a></emu-nt>
  <emu-geq>::</emu-geq>
  <emu-oneof>one of</emu-oneof>
  <emu-rhs>
    <del><emu-t>a</emu-t></del>
    <emu-t>b</emu-t>
    <emu-t>c</emu-t>
    <mark><emu-t>d</emu-t></mark>
    <ins><emu-t>e</emu-t><emu-t>f</emu-t></ins>
  </emu-rhs>
</emu-production>
```
It looks kind of funny (see below), but that's an unrelated issue with how we emit grammar nodes (https://github.com/tc39/ecmarkup/issues/248).

Screen Shot 2020-09-05 at 6 41 47 PM

### DIFF
--- a/src/RHS.ts
+++ b/src/RHS.ts
@@ -35,25 +35,43 @@ export default class RHS extends Builder {
   }
 
   terminalify(parentNode: Element) {
+    // we store effects to perform later so the iteration doesn't get messed up
+    let pairs: { parent: Element; child: Text }[] = [];
     for (let i = 0; i < parentNode.childNodes.length; i++) {
       const node = parentNode.childNodes[i];
-
       if (node.nodeType === 3) {
-        const text = node.textContent!.trim();
-        const pieces = text.split(/\s/);
-
-        pieces.forEach(p => {
-          if (p.length === 0) {
-            return;
+        pairs.push({ parent: parentNode, child: node as Text });
+      } else if (
+        node.nodeType === 1 &&
+        (node.nodeName === 'INS' || node.nodeName === 'DEL' || node.nodeName === 'MARK')
+      ) {
+        for (let i = 0; i < node.childNodes.length; i++) {
+          const child = node.childNodes[i];
+          if (child.nodeType === 3) {
+            pairs.push({ parent: node as Element, child: child as Text });
           }
-          const est = this.spec.doc.createElement('emu-t');
-          est.textContent = p;
-
-          parentNode.insertBefore(est, node);
-        });
-
-        parentNode.removeChild(node);
+        }
       }
     }
+    for (let { parent, child } of pairs) {
+      this.wrapTerminal(parent, child);
+    }
+  }
+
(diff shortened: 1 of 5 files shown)


src/RHS.ts
```diff
@@ -35,25 +35,44 @@ export default class RHS extends Builder {
...
-        const pieces = text.split(/\s/);
-
-        pieces.forEach(p => {
-          if (p.length === 0) {
-            return;
+        pairs.push({ parent: parentNode, child: node as Text });
+      } else if (
+        node.nodeType === 1 &&
+        // @ts-ignore
+        (node.tagName === 'INS' || node.tagName === 'DEL' || node.tagName === 'MARK')
```
rbuckton, line 47: curious what error you're ignoring here. Is it because `node` here is typed `Node` and not `Element`? Can you use `node.nodeName` instead of `node.tagName` here?

bakkot: Yup, it's typed as ChildNode, which lacks `tagName`:

```
src/RHS.ts:46:15 - error TS2339: Property 'tagName' does not exist on type 'ChildNode'.

46         (node.tagName === 'INS' || node.tagName === 'DEL' || node.tagName === 'MARK')
```

and yup, switching to `nodeName` fixes it and works just as well.

Kind of a shame that TypeScript can't use the `nodeName === 1` to refine to `Element`.

rbuckton (approved):

src/RHS.ts
rbuckton: It _could_ if `node`'s type was a discriminated union (i.e. a closed set with `nodeType` as a discriminant property), however `Node` (and `ChildNode`) are open-ended base types and therefore could theoretically be implemented by a value that returns `1` that *isn't* an `Element`.

You *can* however use `instanceof` as a type guard: 
```ts
if (node instanceof Element && node.tagName === 'INS') {
}
```

You can also use a custom type guard:
```ts
function isElement(node: Node): node is Element {
  return node.nodeType === 1;
}
...
if (isElement(node) && node.tagName = ...
```

Its a small enough function that v8 can easily inline.

bakkot: Generally speaking I try not to add code to please the type checker when the code is already well-typed but TypeScript can't prove it; that's what `@ts-ignore` is for, right?

(It's well-typed in the sense that `nodeType` is logically intended to be a discriminant and any code which violates that has violated the type contract, even if that's not a contract TypeScript enforces.)

PR #254, opened 2020-09-08
Fix source text resolution for imported emu-grammar tags

### LINKED ISSUE #253: `emu-grammar` tag content incorrect
While trying to render the spec for https://tc39.es/proposal-class-access-expressions/#sec-left-hand-side-expressions, it seems like ecmarkup is trying to slice the content of the grammar from the node location. However, the node comes from an imported file (via `<emu-import>`) and all of the slices are coming from the root file (index.html), so none of the content is correctly passed.

rbuckton:
This adds a step to walk up the DOM to find a containing `emu-import` to use the `JSDOM` and source text attached to it rather than the root of the spec for `<emu-grammar>` tags in imported elements.

This also fixes some issues in the tests that made it impossible to run the tests cleanly on Windows. The tests _should_ be OS agnostic now.

Fixes #253


bakkot:
You should be able to call [the existing method which does this](https://github.com/tc39/ecmarkup/blob/5b818603458acb247e33b0052e74c0142545b2c9/src/Spec.ts#L428) rather than doing the traversal yourself, I think.

rbuckton:
`locate` works for my case, but fails in one of the tests:
image

rbuckton:
I assume that's what you meant by losing location information?

rbuckton:
Oddly, that's a case where it's *not* imported.

bakkot:
Looks like that case is one where the `emu-grammar` tag is inside of an `emu-alg` tag, so, yeah, that's what I meant about losing location information. Fixing that will probably be annoying.

For the purposes of getting this PR in, then, would you mind factoring out the common parts of your implementation and the existing `locate` method (specifically, the "find me the appropriate `dom` object and source text for this node" part) into a method on `Spec`, and using that method in both places? I'll worry about fixing the `locate` method for the general case later.

I can push a commit which does that if it's not clear.

rbuckton:
The `locate` method itself seems unsafe. It expects `dom.nodeLocation` to always return a value, which is why its throwing above.

rbuckton:
I will have a commit up shortly that makes the result of `locate` possibly missing (i.e. `undefined`) so that it can be correctly checked for in other places.

rbuckton:
How does that handoff work currently? Ecmarkup hands off to ecmarkdown and then what? Ecmarkdown calls grammarkdown or does it send a tree back that Ecmarkup then parses in a separate pass?

rbuckton:
You could theoretically subclass `ecmarkdown.Emitter` to emit DOM nodes and attach location information to them, rather than relying on the string output...

bakkot:
> a commit up shortly that makes the result of `locate` possibly missing

Works for me, thanks.

> Ecmarkdown calls grammarkdown or does it send a tree back that Ecmarkup then parses in a separate pass?

Ecmarkdown sends back [raw HTML](https://github.com/tc39/ecmarkup/blob/5b818603458acb247e33b0052e74c0142545b2c9/src/Algorithm.ts#L44-L48) (not yet processed by grammarkdown) which gets added to the main tree using JSDOM, and then any `emu-grammar` nodes are visited by the [walker](https://github.com/tc39/ecmarkup/blob/5b818603458acb247e33b0052e74c0142545b2c9/src/Spec.ts#L990) just like `emu-grammar` nodes in the original tree.

bakkot:
> You could theoretically subclass ecmarkdown.Emitter to emit DOM nodes and attach location information to them, rather than relying on the string output...

Yup, although a simpler fix would be to invoke JSDOM's parser directly and adopt the resulting nodes, rather than using `.innerHTML =`. That's what [we already do for emu-import](https://github.com/tc39/ecmarkup/blob/5b818603458acb247e33b0052e74c0142545b2c9/src/Import.ts) nodes.

rbuckton:
I suppose that if you used the source text and `JSDOM` from the resulting algorithm, that might work.

bakkot (approved):

PR #267, opened 2020-11-18
use real source text for emu-grammar nodes in main document

### LINKED ISSUE #266: "but only" restriction in grammar does not render
cf https://github.com/tc39/ecma262/issues/570#issuecomment-729349410

cc @rbuckton @jmdyck

bakkot:
Fixes #266.

The issue was that https://github.com/tc39/ecmarkup/pull/254 was inadvertently causing the rendering of `emu-grammar` nodes in the main document to use the `.innerHTML` rather than the real source, meaning it was passing `[...] DecimalEscape [&gt; but only [...]` to grammarkdown. Grammarkdown then silently dropped it. (Which is surprising to me; I'd expect at least an error. cc @rbuckton.)

This fixes the issue from #254 and also makes the fallback path robust against this.

### DIFF
--- a/src/Grammar.ts
+++ b/src/Grammar.ts
@@ -24,7 +24,7 @@ export default class Grammar extends Builder {
     // If the source text is available, we should use it since `innerHTML` serializes the
     // DOM tree beneath the node. This can result in odd behavior when the syntax is malformed
     // in a way that parse5 does not understand, but grammarkdown could possibly recover.
-    if (location && location.source) {
+    if (location) {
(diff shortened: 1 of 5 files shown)


rbuckton:
I'll look at that. Grammarkdown should be interpreting html entities as the character the entity represents.

rbuckton:
Ah, the issue in `grammarkdown` is in the lookahead for `[<`. I'm updating my test in `grammarkdown` and I can fix the handling there.

michaelficarra (approved):

PR #279, opened 2020-12-24
lint: some checks for legal tags/attributes

bakkot:
- lint against unrecognized `emu-` tags
- lint for use of "oldid" (should be "oldids")

~This includes `emu-see-also-para` as a valid tag because it appears in 262, but ecmarkup doesn't know what that is, so I'm probably just going to remove it from 262 and drop it here.~ Edit: https://github.com/tc39/ecma262/pull/2271 drops all occurrences of this tag, so it's no longer included here.

### DIFF
--- a/src/Builder.ts
+++ b/src/Builder.ts
@@ -23,7 +23,7 @@ export default class Builder {
     if (nodeId !== null) {
       if (spec.nodeIds.has(nodeId)) {
         spec.warn({
-          type: 'attr',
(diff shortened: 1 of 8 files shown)


michaelficarra (approved):
Should `attr-value` point to the position of the value or the attribute? I could see either way.

bakkot:
`attr-value` points to the value of the attribute, `attr` points to the attribute itself. Seemed natural to me.

PR #284, opened 2021-01-09
bump highlight.js

bakkot:
Note that the stylesheet is not automatically included, and will need to be bumped in consumers, including ecma262 and @ljharb's [template-for-proposals](https://github.com/tc39/template-for-proposals). Buuuuut no one appears to have updated the stylesheet in the spec at the last major upgrade, so while this is technically a breaking change I don't think it will cause any consumer to be more broken than it currently is. Still, it's probably worth calling it a breaking change.

Fixing the stylesheet gives us a little more highlighting: contrast before

Screen Shot 2021-01-09 at 11 57 32 AM

and after

Screen Shot 2021-01-09 at 11 57 28 AM


ljharb (approved):
the cve doesn’t apply, of course, since we’re not highlighting untrusted user data

bakkot:
Yeah, it's mostly the extremely obnoxious message which gets printed when compiling the spec which I'm worried about.

PR #290, opened 2021-02-12
dynamically add copyable step numbers

bakkot:
As an alternative to https://github.com/tc39/ecmarkup/pull/289, this PR tries to repair the regular copy-paste functionality. From discussion with the other ecma262 editors, we generally prefer this route: in particular the usual case is that you want to copy exactly one line, by triple-clicking it, and this serves that use-case better and without any new UI elements.

I'd [originally done this](https://github.com/tc39/ecmarkup/compare/e81364c..list-numbers) during the build rather than at runtime, but that turns out to make the rendered ecma262 spec 10% (600kb) larger. This seems better. It doesn't seem to noticeably affect time-to-interactive on the browsers I tested on.


js/listNumbers.js
```diff
@@ -0,0 +1,33 @@
+var decimalBullet = Array.apply(null, Array(100)).map(function (a, i) {
```
ljharb, line 1: any reason not to use Array.from here rather than spreading a sparse array?

bakkot: This file is written in ES5. It's probably a good idea to switch over to more modern JS, but I don't want to that as part of this PR.

ljharb: This is a brand new file - do you mean the other files are written in ES5, and you wanted this one to match?

bakkot: Sorry, yes, that.

michaelficarra (approved):

PR #292, opened 2021-02-13
add support for normative optional attribute

bakkot:
cf https://github.com/tc39/ecma262/pull/2215.

Most of the work here was tweaking the existing CSS so that the orange box wouldn't extend up into the whitespace preceding the clause.

The "normative optional" text is now selectable, and if the term "Normative Optional" is defined it will link to that definition.

Before (with ecma262's CSS):

Screen Shot 2021-02-12 at 6 43 27 PM

After (with this PR's CSS):

Screen Shot 2021-02-12 at 6 43 24 PM


ljharb (approved):
nice

michaelficarra (approved):

PR #298, opened 2021-02-20
Modernize JS

bakkot:
I think we can drop support for pre-ES6 browsers now.

I was too lazy to go all the way and convert the ES5-style classes to ES6 `class`es and so on.

Based on https://github.com/tc39/ecmarkup/pull/297; marked as a draft until that lands.

### DIFF
--- a/docs/ecmarkup.js
+++ b/docs/ecmarkup.js
@@ -1026,10 +1026,8 @@ document.addEventListener('DOMContentLoaded', () => {
   referencePane.init();
 });
 'use strict';
-let decimalBullet = Array.apply(null, Array(100)).map((a, i) => '' + (i + 1));
-let alphaBullet = Array.apply(null, Array(26)).map((a, i) =>
-  String.fromCharCode('a'.charCodeAt(0) + i)
-);
+let decimalBullet = Array.from({ length: 100 }, (a, i) => '' + (i + 1));
+let alphaBullet = Array.from({ length: 26 }, (a, i) => String.fromCharCode('a'.charCodeAt(0) + i));
 
 // prettier-ignore
 let romanBullet = ['i', 'ii', 'iii', 'iv', 'v', 'vi', 'vii', 'viii', 'ix', 'x', 'xi', 'xii', 'xiii', 'xiv', 'xv', 'xvi', 'xvii', 'xviii', 'xix', 'xx', 'xxi', 'xxii', 'xxiii', 'xxiv', 'xxv'];
--- a/docs/ecmarkup.js
+++ b/docs/ecmarkup.js
@@ -1,7 +1,6 @@
 'use strict';
-
-var sdoBox = {
-  init: function () {
+let sdoBox = {
+  init() {
     this.$alternativeId = null;
     this.$outer = document.createElement('div');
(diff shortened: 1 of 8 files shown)


js/listNumbers.js
```diff
@@ -1,33 +1,32 @@
-var decimalBullet = Array.apply(null, Array(100)).map(function (a, i) {
-  return '' + (i + 1);
-});
-var alphaBullet = Array.apply(null, Array(26)).map(function (a, i) {
-  return String.fromCharCode('a'.charCodeAt(0) + i);
-});
+'use strict';
+let decimalBullet = Array.apply(null, Array(100)).map((a, i) => '' + (i + 1));
+let alphaBullet = Array.apply(null, Array(26)).map((a, i) =>
```
ljharb, line 3: nit:
```suggestion
let decimalBullet = Array.from({ length: 100 }, (a, i) => '' + (i + 1));
let alphaBullet = Array.from({ length: 26 }, (a, i) =>
```
?

js/listNumbers.js
```diff
@@ -1,33 +1,32 @@
...
-  for (var i = 0; i < ol.children.length; ++i) {
-    var child = ol.children[i];
-    var index = parentIndex.concat([i]);
-    var applicable = bullets[Math.min(index.length - 1, 5)];
-    var span = document.createElement('span');
+  for (let i = 0; i < ol.children.length; ++i) {
+    let child = ol.children[i];
+    let index = parentIndex.concat([i]);
+    let applicable = bullets[Math.min(index.length - 1, 5)];
+    let span = document.createElement('span');
```
ljharb, line 17: none of these are reassigned; can they all be const? (`prefer-const` will autofix these, if that's the style we want)
```suggestion
  for (let i = 0; i < ol.children.length; ++i) {
    const child = ol.children[i];
    const index = parentIndex.concat([i]);
    const applicable = bullets[Math.min(index.length - 1, 5)];
    const span = document.createElement('span');
```

bakkot: Ehhhh I am not really a fan of prefer-const.

docs/ecmarkup.js
```diff
@@ -348,43 +326,40 @@ function Menu() {
...
+  for (let i = 0; i < tocItems.length; i++) {
+    let $item = tocItems[i];
     $item.addEventListener(
       'click',
-      function ($item, event) {
+      (($item, event) => {
         $item.classList.toggle('active');
         event.stopPropagation();
-      }.bind(null, $item)
+      }).bind(null, $item)
```
michaelficarra, line 337: Just close over `$item` now that it's `let`-declared.

js/menu.js
```diff
@@ -254,43 +238,40 @@ function Menu() {
...
+  for (let i = 0; i < tocItems.length; i++) {
+    let $item = tocItems[i];
     $item.addEventListener(
       'click',
-      function ($item, event) {
+      (($item, event) => {
         $item.classList.toggle('active');
         event.stopPropagation();
-      }.bind(null, $item)
+      }).bind(null, $item)
```
michaelficarra, line 249: Same.

michaelficarra (approved):
LGTM otherwise.

PR #299, opened 2021-02-21
move json out of main html

bakkot:
Instead of having an inline `script type="application/json"` for the biblio and the SDO map (which are used for the references pane), put the JSON inline (in a template string) in the generated JavaScript file. That way it can be loaded in parallel with the rest of the page.

~This does have the downside that it becomes possible for the biblio and the HTML to be out of sync, particular given that the JS can be cached. (ECMA-262 is served from github pages, which caches assets for ten minutes.) But since the biblio isn't integral to the document (it exists just to have the references pane work), that seems an acceptable cost.~ Edit: no longer true; thanks Jordan!

Based on https://github.com/tc39/ecmarkup/pull/298; marked as a draft until that lands.


ljharb:
We could mitigate the caching issue by using a query string containing a content hash of the json file.

bakkot:
Oh, good idea. Done.

ljharb (approved):

michaelficarra (approved):

PR #300, opened 2021-02-21
upstream css from ecma262

bakkot:
This pulls in CSS overrides from ecma262.

For the scroll-margin-top CSS added in https://github.com/tc39/ecma262/pull/2317, I've restricted it to clauses, since I noticed it doesn't work so well for other things that lack whitespace above them. cc @ljharb 

For example, here I've navigated to the "It is a Syntax Error if ContainsDuplicateLabels [...]" bullet point.

With the upstream CSS:

Screen Shot 2021-02-20 at 5 41 41 PM

With this PR:

Screen Shot 2021-02-20 at 5 40 49 PM


ljharb (approved):
Maybe sections and annexes also?

bakkot:
Annexes are included. Not sure what you mean by sections?

ljharb:
ah, there's `<section>` elements in older versions of the spec, but i don't see any anymore.

michaelficarra (approved):

PR #301, opened 2021-02-22
multipage builds

### LINKED ISSUE #151: Multipage builds

bakkot:
~Based on https://github.com/tc39/ecmarkup/pull/299; marked as a draft until that lands.~ Edit: ready to go.

Fixes #151.

This is technically a breaking change, in that it in-passing enforces some new constraints on arguments: specifically, you can't specify `--js` when outputting to stdout, since that's nonsensical.

You can see a build of ecma262 rendered using this option [here](https://bakkot.github.io/ecma262-previews/multipage-builds/multipage/). Some notable aspects:

- the reference pane works as expected, and persists when clicking links between sections
- the table of contents works as expected, including pins
- most of the JS is shared between the multipage build and the non-multipage build, to take maximum advantage of https://github.com/tc39/ecmarkup/pull/299
- if you load a page but the fragment is from a different page, you will automatically get redirected
  - there's also an `index.html` so you can treat `/multipage/#foo` as a stable link, if for some reason you want that

I've confirmed that this works much better on Chrome, including on Android, than the single page build.

Things to do in the future, which I didn't want to include in this PR:

- splitting out the table of contents. It's almost half a megabyte of HTML, and it's inlined into every page and hence needs to be loaded every time. Maybe it would be better to stuff the relevant information in a cacheable JSON blob and render it client-side?
  - orrrrrr we could use frames. or load the HTML dynamically and rewrite the URL! Single-Page Apps!
    - I'm not doing those.
- support for emu-import. ecma262 doesn't use it, and it's the primary intended consumer. I will address this at some point.


bakkot:
Should probably add some whitespace after the spec container: https://bakkot.github.io/ecma262-previews/multipage-builds/multipage/conformance.html#sec-conformance.normative-optional

src/Spec.ts
```diff
@@ -550,27 +590,248 @@ export default class Spec {
...
+    const multipageScript = this.doc.createElement('script');
+    multipageScript.src = 'multipage.js?cache=' + sha(multipageJsContents);
+    multipageScript.setAttribute('defer', '');
+    head.appendChild(multipageScript);
+
+    if (!sections.some(({ name }) => name === 'index')) {
+      let placeholder = this.doc.createElement('p');
+      placeholder.innerHTML =
+        'This page serves as an index. If you are not automatically redirected, you can navigate to the actual specification by using the table of contents on the left.';
+      let indexRedrectScript = this.doc.createElement('script');
```
bakkot, line 757:
```suggestion
      let indexRedirectScript = this.doc.createElement('script');
```

michaelficarra (approved):

bakkot:
Google's [page about](https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls) `link rel=canonical` mentions only _duplicate_ pages, not subsets. Search engines are the intended consumer, and it seems like they don't intend it to be used in circumstances like this, so I'm going to leave it out for now, regardless of what the RFC says.

ljharb:
I don't agree. Google's page talks about "similar" pages, of which subsets are arguably included (i'd say objectively, but i suppose it could be debated).

What is the downside to including it?

bakkot:
The downside is that the multipage builds wouldn't show up in search results, which I think is unfortunate: they are often the thing I want. But I'll include it for now and we can back it out later.

PR #304, opened 2021-03-02
multipage: rewrite relative links

bakkot:
Includes version bump commit, so please **rebase**, not squash.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "ecmarkup",
-	"version": "7.0.0",
+	"version": "7.0.1",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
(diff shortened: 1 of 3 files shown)


michaelficarra:
I'd be more comfortable using the other properties like `protocol` and `pathname` instead of testing and slicing `href`.

bakkot:
That's what I started with, but they don't work correctly in jsdom (at least in this version, which we [can't upgrade](https://github.com/jsdom/jsdom/issues/2968)).

michaelficarra:
Okay, fine. Can we extract anything that touches `href` to an appropriately-named function which can be reimplemented in a better way in the future?

bakkot:
Sure? I don't think it's an improvement, but done.

michaelficarra (approved):

PR #307, opened 2021-03-10
Fix ClauseWalker for compatibility with <emu-import>

gibson042:
This is necessary for "update active clause on scroll" functionality on
https://tc39.es/ecma402

### DIFF
--- a/docs/ecmarkup.js
+++ b/docs/ecmarkup.js
@@ -450,20 +450,28 @@ function ClauseWalker(root) {
     NodeFilter.SHOW_ELEMENT,
     {
       acceptNode(node) {
+        // reject nodes under an already-accepted element
         if (previous === node.parentNode) {
           return NodeFilter.FILTER_REJECT;
-        } else {
-          previous = node;
         }
+
+        // descend into <emu-import> to inspect immediate children
+        if (node.nodeName === 'EMU-IMPORT') {
+          return NodeFilter.FILTER_SKIP;
+        }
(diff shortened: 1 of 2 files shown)


bakkot (approved):
Sweet, thanks!

gibson042:
FYI, it's also worth noting that the TreeWalker implementation has about half the efficiency of a simple generator. If you're not married to the former, we can switch to something like
```js
function* getChildClauses(root) {
  for (let el of root.children) {
    switch (el.nodeName) {
      // descend into <emu-import>
      case 'EMU-IMPORT':
        yield * getChildClauses(el);
        break;

      // accept <emu-clause>, <emu-intro>, and <emu-annex>
      case 'EMU-CLAUSE':
      case 'EMU-INTRO':
      case 'EMU-ANNEX':
        yield el;
    }
  }
}
```

bakkot:
Generator works for me. (A lot of this code was written pre-ES6 and I've only put [a little effort](https://github.com/tc39/ecmarkup/pull/298) into modernizing it.)

gibson042:
Let's merge this under-the-covers work first and then follow it up. I got some local errors when developing this that I don't think were my fault, and I'd like them to clear in the main repo before tweaking interfaces.

bakkot:
Works for me.

bakkot:
(or, wait, by "this under-the-covers work" did you mean the generator change? if so, oops.)

PR #309, opened 2021-03-17
Fix formatting: npm run format

gibson042:


bakkot (approved):
Ah, thanks. Gotta make CI enforce this.

PR #310, opened 2021-03-17
Improve findActiveClause performance

gibson042:
Replace TreeWalker-based implementation with a simple generator
Ref https://github.com/tc39/ecmarkup/pull/307#issuecomment-794728029


bakkot (approved):
Thanks! Code looks good and I confirmed locally that it does the right thing.

PR #326, opened 2021-06-16
fixes #325: replace usage of deprecated highlight.js API and bump it

### LINKED ISSUE #325: using deprecated API from highlight.js
When building ecma262, highlight.js outputs the following warning:

> Deprecated as of 10.7.0. highlight(lang, code, ...args) has been deprecated.
> Deprecated as of 10.7.0. Please use highlight(code, options) instead.
> https://github.com/highlightjs/highlight.js/issues/2277

which is triggered by the following line of code:

https://github.com/tc39/ecmarkup/blob/4be89198c095693050e5bd8bf37cd61d9225349a/src/Spec.ts#L1014

We should probably start using the new API to get rid of the warning and allow the highlight.js dependency to be bumped in the future.

michaelficarra:
Fixes #325.


package.json
```diff
@@ -8,7 +8,7 @@
 		"build": "tsc -p src -sourceMap",
 		"build-release": "tsc -p src",
 		"build-spec": "node bin/ecmarkup.js spec/index.html docs/index.html --no-ecma-262-biblio --css-out docs/elements.css --js-out docs/ecmarkup.js",
-		"prepublish": "safe-publish-latest && npm run build-release",
+		"prepack": "safe-publish-latest && npm run build-release",
```
michaelficarra, line 11: I made this change to stop it from running during `npm install`.

package.json
```diff
@@ -69,7 +68,7 @@
 		"mocha": "^5.2.0",
 		"prettier": "^2.0.4",
 		"safe-publish-latest": "^1.1.4",
-		"typescript": "^3.8.2"
+		"typescript": "^4.3.2"
```
michaelficarra, line 71: I had to update this to use `@ts-expect-error` below.

src/Spec.ts
```diff
@@ -39,8 +39,8 @@ import { lint } from './lint/lint';
 import { CancellationToken } from 'prex';
 import type { JSDOM } from 'jsdom';
 
-const DRAFT_DATE_FORMAT = { year: 'numeric', month: 'long', day: 'numeric', timeZone: 'UTC' };
-const STANDARD_DATE_FORMAT = { year: 'numeric', month: 'long', timeZone: 'UTC' };
+const DRAFT_DATE_FORMAT: Intl.DateTimeFormatOptions = { year: 'numeric', month: 'long', day: 'numeric', timeZone: 'UTC' };
+const STANDARD_DATE_FORMAT: Intl.DateTimeFormatOptions = { year: 'numeric', month: 'long', timeZone: 'UTC' };
```
michaelficarra, line 43: Had to add these because of the new TypeScript compiler version.

src/Spec.ts
```diff
@@ -1011,7 +1011,8 @@ ${await utils.readFile(path.join(__dirname, '../js/multipage.js'))}
       const baseIndentRe = new RegExp('^' + baseIndent, 'gm');
       input = input.replace(baseIndentRe, '');
 
-      const result = hljs.highlight(lang, input);
+      // @ts-expect-error the type definitions for highlight.js are broken
```
michaelficarra, line 1014: The types say that there's a default export with a bunch of properties (including `highlight`). In reality, there is no default export, only named exports. TypeScript doesn't allow you to pull named exports it doesn't know about, and using the default export gets you a runtime error that `highlight` is not a property of `undefined`. The types are broken, so we ignore them.

package.json
```diff
@@ -44,9 +44,9 @@
 	"dependencies": {
 		"chalk": "^1.1.3",
 		"ecmarkdown": "^6.0.0",
-		"eslint": "^7.25.0",
+		"eslint": "^7.28.0",
```
michaelficarra, line 47: I had to update the linter and fix a bunch of linting issues because it was incompatible with the new TypeScript version.

bakkot (approved):

PR #327, opened 2021-06-29
Support for parsing structured headers for AOs and similar operations

bakkot:
As in https://github.com/tc39/ecma262/pull/545.

Headers look like this:

```html
<emu-clause id="sec-exampleao" type="abstract operation">
  <h1>
    ExampleAO (
      param : an integer,
      optional param2 : a String,
    )
  </h1>
  <dl class='header'>
    <dt>description</dt>
    <dd>It is an example.</dd>
  </dl>
  [...]
```

and produce HTML that looks like this (modulo formatting and other details):
```html
<emu-clause id="sec-exampleao" aoid="ExampleAO">
  <h1>ExampleAO ( param [ , param2 ] )</h1>
  <p>The abstract operation ExampleAO takes argument param (an integer) and optional argument param2 (a String). It is an example.</p>
  [...]
```

Of note:
- the name of the AO now appears in the original source once, instead of three times
- the name of each parameter now appears in the original source once, instead of twice
- there is a new `dd` element which we can stuff arbitrary other machine-readable attributes into

This is not a breaking change, since the previous format is still supported.


src/Clause.ts
```diff
@@ -51,16 +52,85 @@ export default class Clause extends Builder {
       // <emu-clause id=foo aoid> === <emu-clause id=foo aoid=foo>
       this.aoid = node.id;
     }
+
+    let header = this.node.firstElementChild;
+    while (header != null && header.tagName === 'SPAN' && header.children.length === 0) {
+      // skip oldids
+      header = header.nextElementSibling;
+    }
+    if (header == null || header.tagName !== 'H1') {
```
michaelficarra, line 61: Missing header and header being the wrong tag are kinda different, and I'd prefer to receive a different message for those cases.

src/Clause.ts
```diff
@@ -51,16 +52,85 @@ export default class Clause extends Builder {
...
+    // if we find such a DL, treat this as a structured header
+
+    const type = this.node.getAttribute('type') ?? 'unknown';
+    this.node.removeAttribute('type'); // TODO maybe leave it in; this is just to minimize the diff
+
+    const { name, formattedHeader, formattedParams } = parseStructuredHeaderH1(this.spec, header);
+    if (type === 'numeric method' && name != null && !name.includes('::')) {
+      this.spec.warn({
+        type: 'contents',
+        ruleId: 'numberic-method-for',
```
michaelficarra, line 89:
```suggestion
        ruleId: 'numeric-method-for',
```

src/Clause.ts
```diff
@@ -51,16 +52,85 @@ export default class Clause extends Builder {
...
+        message: `numeric methods should be of the form \`Type::operation\``,
```
michaelficarra, line 90:
```suggestion
        message: 'numeric methods should be of the form `Type::operation`',
```

src/Clause.ts
```diff
@@ -51,16 +52,85 @@ export default class Clause extends Builder {
...
-    this.header.insertBefore(numElem, this.header.firstChild);
+    const { description, for: _for } = parseStructuredHeaderDl(this.spec, type, dl);
+
+    const paras = formatPreamble(
+      this.spec,
+      this.node,
+      dl,
+      type,
+      name ?? 'UNKNOWN',
+      formattedParams ?? 'UNPARSEABLE ARGUMENTS',
```
michaelficarra, line 108: If these are null, why not just fail?

bakkot: Mostly ecmarkup tries hard to produce some output unless you pass `--strict`. That makes it easier to iterate on the stuff you are actually trying to work on and fix issues in whatever order you want.

src/header-parser.ts
```diff
@@ -0,0 +1,336 @@
...
+    // if it's multiline, parse it for types
+    const paramChunks = paramText
+      .split('\n')
+      .map(c => c.trim())
+      .filter(c => c.length > 1);
+    let index = 0;
+    for (const chunk of paramChunks) {
+      ++index;
+      // TODO linter enforces all optional params come after non-optional params
+      const parsedChunk = chunk.match(/^(optional\s*)?([A-Za-z0-9_]+)\s*:\s*(\S.*\S)/);
```
michaelficarra, line 47:
```suggestion
      const parsedChunk = chunk.match(/^(optional\s+)?([A-Za-z0-9_]+)\s*:\s*(\S.*\S)/);
```

to allow mandatory parameters whose name starts with "optional".

ljharb: Is that a thing we’d actually want to do?

src/header-parser.ts
```diff
@@ -0,0 +1,336 @@
...
+      if (parsedChunk == null) {
+        spec.warn({
+          type: 'contents',
+          ruleId: 'header-format',
+          message: `failed to parse header (parameter ${index})`,
+          node: header,
+          nodeRelativeColumn: 1,
+          nodeRelativeLine: 1,
```
michaelficarra, line 55: We could pretty easily point to the correct line here by zipping the split/trim result with the naturals. I think it might be pretty useful since this is possibly quite a bit of text.

src/header-parser.ts
```diff
@@ -0,0 +1,336 @@
...
+        });
+        continue;
+      }
+      const optional = parsedChunk[1] != null;
+      const paramName = parsedChunk[2];
+      let paramType = parsedChunk[3];
+      if (paramType.endsWith(',')) {
+        paramType = paramType.slice(0, -1);
+      }
+      (optional ? optionalParams : params).push({
```
michaelficarra, line 65: We could warn about non-optional parameters after optional ones pretty easily right here by doing

```js
if (!optional && optionalParams.length > 0) { spec.warn(/* ... */); }
```

src/header-parser.ts
```diff
@@ -0,0 +1,336 @@
...
+        name: paramName,
+        type: paramType === 'unknown' ? null : paramType,
+      });
+    }
+    const formattedprefix = prefix == null ? '' : prefix[0].trim() + ' ';
```
michaelficarra, line 70:
```suggestion
    const formattedPrefix = prefix == null ? '' : prefix[0].trim() + ' ';
```

src/header-parser.ts
```diff
@@ -0,0 +1,336 @@
...
+function eat(text: string, regex: RegExp) {
+  const match = text.match(regex);
+  if (match == null) {
+    return { success: false, match, text };
+  }
+  return { success: true, match, text: text.substring(match[0].length) };
+}
+
+function formatEnglishList(list: Array<string>) {
+  if (list.length === 1) {
```
michaelficarra, line 329: Maybe throw for `list.length === 0`.

src/header-parser.ts
michaelficarra: It's a thing we'd actually want the tool to support. If it didn't, it'd be a confusing (and importantly silent) difference in behaviour from what was written.

src/header-parser.ts
```diff
@@ -0,0 +1,336 @@
...
+    }
+    formattedHeader += ' )';
+  } else {
+    let optional = false;
+    while (true) {
+      if (paramText.length == 0) {
+        break;
+      }
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      let { success, text, match } = eat(paramText, /^\s*\[(\s*,)?/);
```
michaelficarra, line 86: A more elegant solution might just surround each of these `eat` calls (and the successive statements) in blocks so you don't have to worry about declaring an unused var here.

michaelficarra (approved):
LGTM otherwise.

src/Clause.ts
michaelficarra: Well surely we should output a warning in this case, right?

bakkot: [We do](https://github.com/tc39/ecmarkup/blob/7837fc12c4741637e44d7b892e4a7466b46db4dc/src/header-parser.ts#L28), at the point at which parsing fails. The linked line is the only path where `parseStructuredHeaderH1` can return `null`. (I can tweak the structure of that method to make that more obvious, I suppose.)

src/Clause.ts
bakkot: Done.

src/Clause.ts
bakkot: Done.

src/Clause.ts
bakkot: Done.

src/Clause.ts
bakkot: Made the above tweak.

src/header-parser.ts
bakkot: Nice catch, done.

src/header-parser.ts
bakkot: It's more complicated than that, but done.

src/header-parser.ts
bakkot: Done.

src/header-parser.ts
bakkot: Done.

src/header-parser.ts
bakkot: Done. Would be nice if the type system could handle that, but I appreciate the difficulty involved.

src/header-parser.ts
bakkot: Ehh, I'm not going to worry about it. I kind of like all calls to `eat` having the same form.

PR #328, opened 2021-06-29
Use multiple highlight colors for variables.

Ms2ger:
This makes life a lot easier if you want to highlight multiple variables in a single algorithm.

Inspired by the similar [bikeshed](https://github.com/tabatkins/bikeshed/) feature. Bikeshed also has a feature to keep the same color when you unselect and then reselect a variable; I didn't port that.


michaelficarra:
FYI, these are the new highlight colours, in order:

image

And this is the current one:

image

@Ms2ger Where did this colour list come from? Why change the primary one?

michaelficarra (approved):
implementation LGTM

Ms2ger:
Colors also come from bikeshed, with the following comment:
```python
    # Colors were chosen in Lab using https://nixsensor.com/free-color-converter/
    # D50 2deg illuminant, L in [0,100], a and b in [-128, 128]
    # 0 = lab(85,0,85)
    # 1 = lab(85,80,30)
    # 2 = lab(85,-40,40)
    # 3 = lab(85,-50,0)
    # 4 = lab(85,5,15)
    # 5 = lab(85,-10,-50)
    # 6 = lab(85,35,-15)
```

I'm happy to go any way you prefer.

michaelficarra:
I don't think these colours are readable enough as-is. Compare the current

image

to the proposed primary colour

image

The secondary colour is even more egregious:

image

I think we should either choose a different set of colours that contrast well against our current alias text colour (`#2AA198`) or include a contrasting text colour in each of the `referenced` classes. If we want to go the former route, it looks like this tool does exactly what we'd want: https://learnui.design/tools/accessible-color-generator.html. If we go the latter route, I'd prefer to limit the text colour to white(-ish) or black(-ish).

Ms2ger:
Oh right, I forgot that bikeshed uses plain black for variables. That seems fine for all of them:
Background colors

If you'd rather change the background colors, I'd prefer if you could come up with a list.

bakkot:
I've pushed a couple of commits which 1) make variables use the default almost-black text color when highlit, which addresses the contrast concern and b) restores the original color for the first variable, since I think it has adequate contrast with the other colors and hence does not require changing (and potentially confusing readers about why it's a different color than they're used to).

I think both changes look fine, personally, and I don't really want to try to find 8 colors which contrast well with the our variables' default greenish color. (The tool linked above produces very pale, frequently similar colors when asked to find variations of this set of eight colors which contrast adequately with `#2AA198`.) @michaelficarra thoughts?

Screen Shot 2021-07-04 at 4 44 09 PM

michaelficarra:
Works for me, though I do think white contrasts better with some of the chosen colours than black. But it's probably better to just be consistent with black text. Either way is fine for me.

michaelficarra (approved):

bakkot:
Thanks @Ms2ger!

Ms2ger:
Thanks for the help!

PR #329, opened 2021-07-04
transfer nodes rather than going via innerHTML

bakkot:
Using `.innerHTML =` recreates nodes, which means we lose track of their source-location information (which is stored in a Map). Instead, just move the original nodes.


src/Example.ts
```diff
@@ -40,7 +40,11 @@ export default class Example extends Builder {
       });
     }
 
-    this.node.innerHTML = '<figure>' + this.node.innerHTML + '</figure>';
+    const ele = this.spec.doc.createElement('figure');
+
+    // @ts-ignore childNodes is iterable
+    ele.append(...this.node.childNodes);
+    this.node.append(ele);
```
michaelficarra, line 47: Won't the old one replace all current children and the new one keep all current children?

bakkot: No, `ele.append(...this.node.childNodes);` removes all children of `node`. When have A with child B and then you append B to a new node C, B is removed from A's children.

michaelficarra (approved):

PR #334, opened 2021-07-16
deps: add parse5 to devDependencies

lightmare:
`parse5` is imported from `src/Spec.ts` and `src/utils.ts` but was missing from `devDependencies`.

I used `pnpm` to install deps, which doesn't lift transitive dependencies like `npm` does.


bakkot (approved):

bakkot:
Thanks!

PR #335, opened 2021-07-16
typescript: use DOM.Iterable type library

lightmare:
Enable `dom.iterable` in `tsconfig` and remove a few unnecessary `ts-ignore` comments, `as Iterable<Element>` assertions, and `Array.from` calls.

### DIFF
--- a/src/Spec.ts
+++ b/src/Spec.ts
@@ -716,7 +716,6 @@ export default class Spec {
         contained.push(child.id);
         containedIdToSection.set(child.id, name);
 
-        // @ts-ignore
         for (const item of child.querySelectorAll('[id]')) {
           contained.push(item.id);
           containedIdToSection.set(item.id, name);
@@ -974,7 +973,7 @@ ${await utils.readFile(path.join(__dirname, '../js/multipage.js'))}
   private async loadBiblios() {
     this.cancellationToken.throwIfCancellationRequested();
     await Promise.all(
-      Array.from(this.doc.querySelectorAll('emu-biblio')).map(biblio =>
+      Array.from(this.doc.querySelectorAll('emu-biblio'), biblio =>
         this.loadBiblio(path.join(this.rootDir, biblio.getAttribute('href')!))
       )
     );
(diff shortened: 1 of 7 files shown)


src/header-parser.ts
```diff
@@ -321,15 +317,13 @@ export function formatPreamble(
     }
   }
   if (description != null) {
-    // @ts-ignore childNodes is iterable
-    const isJustElements = [...(description.childNodes as Iterable<Node>)].every(
+    const isJustElements = Array.prototype.every.call(
```
bakkot, line 320: Can you revert this change? As a matter of style I prefer spreading and calling array methods directly over using `.call`.

src/Spec.ts
```diff
@@ -1243,10 +1236,13 @@ ${await utils.readFile(path.join(__dirname, '../js/multipage.js'))}
         });
         continue outer;
       }
+      if (!header) {
```
bakkot, line 1239: This is unreachable; please remove it.

src/Spec.ts
```diff
@@ -968,20 +967,20 @@ ${await utils.readFile(path.join(__dirname, '../js/multipage.js'))}
...
-    await this.loadBiblio(path.join(__dirname, '../ecma262biblio.json'));
+    await this.loadBiblio(__dirname, '../ecma262biblio.json');
   }
 
   private async loadBiblios() {
     this.cancellationToken.throwIfCancellationRequested();
     await Promise.all(
-      Array.from(this.doc.querySelectorAll('emu-biblio')).map(biblio =>
-        this.loadBiblio(path.join(this.rootDir, biblio.getAttribute('href')!))
+      Array.prototype.map.call(this.doc.querySelectorAll('emu-biblio'), biblio =>
```
bakkot, line 976: As below, please revert this change.

src/Spec.ts
```diff
@@ -968,20 +967,20 @@ ${await utils.readFile(path.join(__dirname, '../js/multipage.js'))}
...
+        this.loadBiblio(this.rootDir, biblio.getAttribute('href')!)
       )
     );
   }
 
-  private async loadBiblio(file: string) {
-    const contents = await this.fetch(file);
+  private async loadBiblio(...file: string[]) {
```
bakkot, line 982: Please revert this change; it's not related, and I prefer the original.

src/header-parser.ts
lightmare: That's forfeiting the possibility of not exhausting the whole list if every short-circuits; but okay, it wouldn't really buy anything here, anyway.

lightmare:
Amended with loadBiblio and Array.prototype.every reverted.

bakkot (approved):

bakkot:
Thanks!

PR #336, opened 2021-07-18
do not copy links into the table of contents

### LINKED ISSUE #333: Links in headers should be removed in the TOC
In Temporal, we have a bunch of sections that override 262/402 sections, and we link to the original in the header of the section. Unfortunately, that leads to TOC markup like

```html
<span class="item-toggle">◢</span>
<a href="#sup-time-zone-names" title="Time Zone Names"><span class="secnum">15.1</span> </a>
<a href="https://tc39.es/ecma402/#sec-time-zone-names">Time Zone Names</a>
```

which means that clicking the title in the TOC brings you to the original 402 prose rather than the override.

bakkot:
Fixes #333.


src/Clause.ts
```diff
@@ -201,7 +201,12 @@ export default class Clause extends Builder {
       clause.title = 'UNKNOWN';
       clause.titleHTML = 'UNKNOWN';
     } else {
-      ({ textContent: clause.title, innerHTML: clause.titleHTML } = header);
+      const headerClone = header.cloneNode(true) as Element;
+      for (const a of headerClone.querySelectorAll('a')) {
+        a.replaceWith(...a.childNodes);
+      }
+      clause.titleHTML = headerClone.innerHTML;
+      clause.title = header.textContent;
```
michaelficarra, line 209:
```suggestion
      clause.title = headerClone.textContent;
```

michaelficarra (approved):

src/Clause.ts
bakkot: Should be equivalent, but sure, done.

PR #337, opened 2021-07-18
add a snapshot of a multipage build

bakkot:
The snapshot lives in a directory named `multipage.html`, which is a bit odd but I don't care enough to special-case it.

Third commit isn't really related to this PR, I just noticed it in passing and figured it would be a good demonstration of the new snapshot.


test/baselines.js
```diff
@@ -23,29 +25,47 @@ function build(file) {
...
+
+      let threw = true;
+      try {
+        if (expectedFiles.size === 0) {
+          throw new Error('Baseline does not exist');
+        }
+        assert.deepStrictEqual(actualFiles, expectedFiles);
+        threw = false;
+      } finally {
+        if (threw || true) {
```
michaelficarra, line 62: ?

bakkot: Whoops, left over from testing it. Fixed.

michaelficarra (approved):

PR #341, opened 2021-07-22
cleanup some mess I left in autolink

lightmare:
Follow-up to #340, implements some changes suggested by @gibson042 

I've replaced my dubious implementation of `longestCommonPrefix` with a **KISS** alternative. The binary search variant was 1) kinda overkill, and 2) poorly implemented — bad estimate of max length at the beginning, and then repeatedly comparing whole prefixes including the parts that were already known to be equal across the whole array.

The new linear algorithm is hopefully much easier to understand, and quite possibly even more efficient for the use case expecting very short prefixes. Strictly speaking it's *O(result.length * items.length)*, which looks quadratic, but that product is always <= `items.join().length`, so the worst-case complexity is linear with respect to the total size of input in characters.

### DIFF
--- a/src/autolinker.ts
+++ b/src/autolinker.ts
@@ -107,20 +107,15 @@ function isCommonAbstractOp(op: string) {
 }
 
 function lookAheadBeyond(entry: BiblioEntry) {
-  if (entry.type === 'term') {
-    if (/^\w/.test(entry.key!)) {
-      return '\\b(?!\\.\\w|%%|\\]\\])';
-    } else {
-      return '';
-    }
-  } else {
-    // type is "op"
-    if (isCommonAbstractOp(entry.key!)) {
-      return '\\b(?=\\()';
-    } else {
-      return '\\b(?!\\.\\w|%%|\\]\\])';
-    }
+  if (entry.type === 'op' && isCommonAbstractOp(entry.key!)) {
+    // must be followed by parentheses
+    return '\\b(?=\\()';
   }
+  if (entry.type !== 'term' || /^\w/.test(entry.key!)) {
+    // must not be followed by `.word` or `%%` or `]]`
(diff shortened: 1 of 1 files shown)


src/autolinker.ts
```diff
@@ -144,40 +142,41 @@ function regexpUnion(alternatives: string[]) {
...
-    result = prefix;
+    ++endIndex;
   }
-  return result;
+  return items[0].slice(beginIndex, endIndex);
 }
 
-function regexpForList(autolinkmap: AutoLinkMap, autolinkkeys: string[], position: number) {
+function regexpPatternForAutolinkKeys(
+  autolinkmap: AutoLinkMap,
```
lightmare, line 163: I did not rename this one to `biblio` as suggested, because that name is used elsewhere for something completely different, while this thing is named all-lowercase `autolinkmap` everywhere it's used.

src/autolinker.ts
```diff
@@ -107,20 +110,15 @@ function isCommonAbstractOp(op: string) {
...
-    } else {
-      return '\\b(?!\\.\\w|%%|\\]\\])';
-    }
+  if (entry.type === 'op' && isCommonAbstractOp(entry.key!)) {
+    // must be followed by parentheses
+    return '\\b(?=\\()';
+  }
+  if (entry.type !== 'term' || /^\w/.test(entry.key!)) {
+    // must not be followed by `.word` or `%%` or `]]`
+    return '\\b(?!\\.\\w|%%|\\]\\])';
```
lightmare, line 119: I went with comment re-phrasing what the regexp says, instead of putting it in an appropriately named variable. The regexp is pretty dense, so having it spelled out can be helpful.

gibson042 (approved):
Thanks!

bakkot (approved):

PR #345, opened 2021-08-18
Add `generatedFiles` to public API for `Spec`

### LINKED ISSUE #344: Missing `generatedFiles` in public API
In #301, the `toHTML()` method was removed from the public API, which I was using in `gulp-emu`. I was just updating `gulp-emu` to support the latest ecmarkup, but `generatedFiles` was never added as a replacement, so there's no type-safe way to access the generated outputs.

rbuckton:
Fixes #344

### DIFF
--- a/src/Spec.ts
+++ b/src/Spec.ts
@@ -91,7 +91,7 @@ export default interface Spec {
   rootDir: string;
   namespace: string;
   exportBiblio(): any;
-  generatedFiles: ReadonlyMap<string | null, string>;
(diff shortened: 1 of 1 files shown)


bakkot:
Thanks! I forgot I needed to update this as well as the class itself.

(Looks like TypeScript is complaining about the types, so I can't merge it yet; I can try to fix that myself later if you don't get to it.)

bakkot (approved):

PR #346, opened 2021-08-19
Plural dfns

### LINKED ISSUE #70: Autolinking should account for plurals
If I do `<dfn>early error</dfn>`, then the phrase "early errors" should autolink to that dfn.

This should be a fun one :)

### LINKED ISSUE #157: plural <dfn>s
In the spec, "Module Record", for example, has a `<dfn>`, so all occurrences of `Module Record` are auto-linked, which is great.

However, occurrences of `Module Records` are not.

It probably wouldn't be ideal to silently pluralize dfns, but perhaps something like `<dfn id="whatever" plural="Module Records">Module Record</dfn>` would work?

bakkot:
Fixes #70, fixes #157.

This is arguably a breaking change because it modifies the format of the biblio, though in practice I don't think the modification will break anything. Still, version number are free, might as well call it major.


spec/index.html
```diff
@@ -190,6 +190,23 @@ <h2>Attributes</h2>
...
 
+<emu-clause id="definitions">
+  <h1>Definitions</h1>
+  <p>Terms can be defined using the `&lt;dfn>` element. Any uses of that term will be automatically linked to the clause containing the definition, or, if the `&lt;dfn>` element has an `id`, to the `&lt;dfn>` itself. This can be suppressed with the <a href="#emu-not-ref">`emu-not-ref`</a> element.</p>
+  <p>When the term starts with a lowercase English letter, usages of the term with the first letter capitalized will also link.</p>
+  <p>Other variants of the term, such as plural forms, can be provided as a comma-seperated list in the `variants` attribute on the element.</p>
+
+  <h2>Example</h2>
+  <pre><code class="language-html">
+  &lt;p>A &lt;dfn id="swordfish" variants="swordfishes,Xiphias gladius">swordfish&lt;/dfn> is large fish characterized by a long, pointed bill.&lt;/p>
```
michaelficarra, line 201:
```suggestion
  &lt;p>A &lt;dfn id="swordfish" variants="swordfishes,broadbill,broadbills,Xiphias gladius">swordfish&lt;/dfn> is a large fish characterized by a long, pointed bill.&lt;/p>
```

michaelficarra (approved):

PR #357, opened 2021-09-11
Make update-baselines script faster

bakkot:
Rather than shelling out once for each baseline, which has a lot of overhead, just teach the existing baseline runner to recognize an environment variable indicating it should overwrite the reference sources.

This abuses the fact that `npm run` will forward all `--args` as environment variables prefixed with `npm_config_`, even those it doesn't recognize. This works even on windows (I believe), where setting an environment variable with the `VAR=val command` syntax would not.

On my machine, this cuts the wall-clock time `npm run update-baselines` takes from 26.7 seconds to 2.5 seconds.

As a bonus, it means that the `update-baselines` script is guaranteed to treat metadata exactly like the test itself does. Previously some options were not treated the same: in particular, setting `ecma262Biblio: false` in the metadata was ignored in the `update-baselines` script but not in the test itself, which meant it was unusable in baselines.


test/baselines.js
```diff
@@ -24,6 +24,8 @@ function build(file) {
 }
 
 describe('baselines', () => {
+  let rebaseline = !!process.env.npm_config_update_baselines;
```
ljharb, line 27: does this still work in npm 7? i'm not sure this is something to rely on, as opposed to `cross-env ENV_VAR=value npm run test-baselines`

bakkot: It does, yes. It's a [documented feature](https://docs.npmjs.com/cli/v7/using-npm/config#command-line-flags); it seems fine to rely on to me - better than relying on a new external dependency, certainly.

ljharb: cross-env is ancient and universal, but if it’s documented then that’s clearly fine.

PR #358, opened 2021-09-11
Add support for structured headers for SDOs

bakkot:
Followup to #327 to extend support to SDOs. As in that PR, the old syntax is still parsed and treated just as today, so this is not a breaking change.

Unlike in #327, the HTML generated here is not identical to what's currently in 262. Specifically, 262 includes either no text at all, the text `With parameters _foo_ and _bar_.`, or on a couple of occasions a paragraph of text. When using structured headers, the output will instead look like

> The syntax-directed operation Name takes parameter _foo_. [optional description] It is defined piecewise over the following productions:

---

There's a slight oddity here: when using this style to specify parameters, you'd write e.g.

```
  <emu-clause id="sec-example" type="sdo">
    <h1>
      Example (
        _param_: an ECMAScript language value,
      )
    </h1>
    <dl class='header'>
    </dl>
```

However, the parameter list never appears in `( list )` form in the output, only in the "takes argument _param_" prose. I think that's fine, but it might be a bit confusing.


jmdyck:
Why not include the parameter 'type' in the generated prose?

bakkot:
> Why not include the parameter 'type' in the generated prose?

It is included. See e.g. [EvaluateAsyncFunctionBody](https://bakkot.github.io/ecma262-previews/sdo-structured-header/multipage/ecmascript-language-functions-and-classes.html#sec-runtime-semantics-evaluateasyncfunctionbody).

jmdyck:
Then I don't understand the "slight oddity" you were describing.

PR #362, opened 2021-09-16
Support "effects" in AO structured headers

syg:
This PR adds support for the `effects` `<dt>` key to the AO structured header. The corresponding `<dd>` is a comma-separated list of strings.

The list of effects are propagated to all clauses that have an Xref to (i.e. users of) the AO with the effect, transitively. Some effects are runtime-only and do not propagate to static SDOs. Currently, only `user-code` (i.e. "can call user code") is included in that list.

During autolinking, linked AOs and SDOs that are invocations (current supported styles are `AO()` and `SDO of`) with effects get the effect names as class names prefixed with `e-` on the `<a>` tag. For example, if a link to an AO has the `user-code` "can call user code" effect, links to it would include the class `e-user-code` in applicable contexts. For `user-code`, since that only applies at runtime, that class doesn't appear in links within static SDO clauses.

Since the propagation of effects is conservative, there may be false positives. The list of effects may be suppressed with the `<emu-meta>` tag around the AO invocation: `<emu-meta suppress-effects="user-code">AOName()</emu-meta>`.


syg:
Forgot to add that there is preliminary styling for `e-uc`, which can be toggled with the `u` key.

jmdyck:
"uc" seems rather cryptic. Does the effect name have to be so short?

> The list of effects may be overridden with the <emu-meta> tag around the AO name

In an invocation, you mean?

syg:
> "uc" seems rather cryptic. Does the effect name have to be so short?

Nope, I'm open to any name here. Originally it was 'can-call-user-code' which was kind of long.

> In an invocation, you mean?

Yes, exactly.

syg:
Also to be clear, I'm not wedded to the way "UC" is styled right now, it's just what I was using to test.

src/Clause.ts
```diff
@@ -180,6 +196,16 @@ export default class Clause extends Builder {
     }
   }
 
+  isEffectApplicable(effectName: string) {
+    // The following effects are runtime only:
+    //
+    // uc: Only runtime can call user code.
+    if (this.type === 'static sdo') {
```
bakkot, line 203: Nothing currently has this type, and the logic [elsewhere](https://github.com/tc39/ecmarkup/blob/a654534daa79e5af43987203f81bfa1c6e726831/src/header-parser.ts#L325-L360) doesn't support it. If the intention is just to exclude "Static Semantics" SDOs, we could track the "Static Semantics" part as an attribute separately from the type.

bakkot:
Looks reasonable at a glance, though we should sort out whether we're going to take generic a generic "effects" attribute or to have a concrete list.

Also needs tests/docs, which I can get to at some point if @syg doesn't.

src/Clause.ts
syg: Ah that's right, I was testing off a local copy of ecma262 where I had annotated static SDOs. I'll make a PR to ecma262.

syg: Switched to using "Static Semantics" prefix.

src/header-parser.ts
```diff
@@ -260,6 +261,13 @@ export function parseStructuredHeaderDl(
         }
         break;
       }
+      case 'effects': {
+        // The dd contains a comma-separated list of effects.
+        if (dd.textContent !== null) {
+          effects = validateEffects(spec, dd.textContent.split(','), dd);
```
bakkot, line 267:
```suggestion
          effects = validateEffects(spec, dd.textContent.split(',').map(c => c.trim()), dd);
```

bakkot (approved):

PR #367, opened 2021-11-02
auto-formatter

### LINKED ISSUE #317: formatter
As I add more lint rules, I increasingly feel the lack of auto-fixing. And thinking about that problem, I think it's time to approach from the other direction: all whitespace-related rules, as well things like capitalization of tags and commas at the end of `1. Repeat,` and so on, would be subsumed by a tool which took in ecmarkup and spit out a canonical formatting. The linter could just run this and assert that the output was identical to the input, like the prettier plugin for eslint.

bakkot:
Fixes #317. You can see this applied to ecma262 at https://github.com/tc39/ecma262/pull/2564.

Still needs docs.

This probably doesn't really need a comprehensive review. I'd just look at `test/formatter.js` to confirm that the things it's enforcing look correct.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "ecmarkup",
-	"version": "9.4.1",
+	"version": "9.5.0",
 	"description": "Custom element definitions and core utilities for markup that specifies ECMAScript and related technologies.",
 	"main": "lib/ecmarkup.js",
 	"typings": "lib/ecmarkup.d.ts",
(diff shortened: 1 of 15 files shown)


bakkot:
Depends on https://github.com/tc39/ecmarkdown/pull/89 getting landed/released, hence the CI failure.

package.json
```diff
@@ -48,7 +48,7 @@
 	"dependencies": {
 		"chalk": "^1.1.3",
 		"dedent-js": "^1.0.1",
-		"ecmarkdown": "../ecmarkdown/ecmarkdown-6.0.1.tgz",
+		"ecmarkdown": "6.0.2",
```
ljharb, line 51:
```suggestion
		"ecmarkdown": "^6.0.2",
```

michaelficarra (approved):

PR #376, opened 2021-11-21
switch from nomnom to command-line-args

### LINKED ISSUE #348: Dependent package `nomnom` is deprecated and has vulnerabilities.
Per `npm audit`:

image

The package [`nomnom`](https://www.npmjs.com/package/nomnom) is no longer maintained and won't receive further updates:. While the likelihood of exploit is low when using `ecmarkup` from the command line, it is much higher when using `ecmarkup` as a library.

Please consider switching to another argument parser.

bakkot:
Fixes #348.

This is a breaking change because nomnom allowed you to pass arbitrary other arguments without failing, whereas this library will exit with an error (which is much more user-friendly). I have some other breaking changes I want to get in at the same time as this, so I'm going to leave this open until I get those done.

### DIFF
--- a/src/arg-parser.ts
+++ b/src/arg-parser.ts
@@ -1,0 +1,103 @@
+import type { OptionDefinition } from 'command-line-usage';
+import * as commandLineArgs from 'command-line-args';
+
+function fail(msg: string): never {
+  console.error(msg);
+  process.exit(1);
+}
+
+export function parse<T extends readonly OptionDefinition[]>(
+  options: T,
+  printHelp: () => void
+): ArgsFromOptions<T> {
+  const def = options.find(o => o.defaultOption);
+  if (!def?.multiple) {
+    // this is just so I don't have to think about how to handle `--` in other cases
+    // not an inherent limitation
+    throw new Error('ecmarkup arg-parser requires a default option for now');
+  }
+
+  const argv = process.argv.slice(2);
+
+  let notParsed: string[] = [];
+  const dashDashIndex = argv.indexOf('--');
+  if (dashDashIndex !== -1) {
+    notParsed = argv.splice(dashDashIndex + 1);
+    argv.pop();
+  }
+
+  let args: ArgsFromOptions<T>;
+  try {
+    // @ts-ignore the types are wrong about mutability
+    args = commandLineArgs(options, { argv });
+  } catch (e: any) {
+    if (e?.name === 'UNKNOWN_OPTION') {
+      fail(`Unknown option ${e.optionName}`);
+    }
+    throw e;
+  }
+
+  // @ts-ignore it's fine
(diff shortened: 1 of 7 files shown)


ljharb:
Why this one instead of the most used ones, yargs or commander?

bakkot:
I've looked at both of those, but both have bad enough APIs that I wasn't willing to use them. Writing my own would be less painful.

ljharb:
Hm, i use yargs in all my things; i love its api.

bakkot:
Yargs seems to be really focused on commands with multiple subcommands, such that simple problems like "this command has one required positional argument and several optional flags" are not trivial to specify. Maybe I'm missing something obvious, but I genuinely could not figure out how to specify a required positional argument except by crafting a special string with multiple pieces of magic syntax to pass to `.command`.

Plus I couldn't get the types to work for TypeScript for required optional arguments.

ljharb:
You've definitely nailed the major thing I think yargs needs to improve; i've been bitten by it before. I don't use TS tho, but you're right that the imperative subcommand stuff would probably not work well with the type inference.

In general I'd still prefer to use a 65M/week very well-supported tool over a 1M/week tool, regardless of a few API quirks, but since ecmarkup has a very finite number of consumers that probably doesn't matter here.

src/arg-parser.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export function parse<T extends readonly OptionDefinition[]>(
+  options: T,
+  printHelp: () => void
+): ArgsFromOptions<T> {
+  const def = options.find(o => o.defaultOption);
+  if (!def?.multiple) {
+    // this is just so I don't have to think about how to handle `--` in other cases
+    // not an inherent limitation
+    throw new Error('ecmarkup arg-parser requires a default option for now');
```
michaelficarra, line 17: This error doesn't help the end user resolve the issue. What are they supposed to do in response to this? Or will this not actually occur in practice?

src/arg-parser.ts
```diff
@@ -0,0 +1,103 @@
...
+    (argv.length === 0 && notParsed.length === 0)
+  ) {
+    printHelp();
+    process.exit(0);
+  }
+
+  return args;
+}
+
+// here follows some typescript shenangans for infering the type of of the parsed arguments from the argument specification
```
michaelficarra, line 54:
```suggestion
// here follows some typescript shenanigans for inferring the type of of the parsed arguments from the argument specification
```

src/arg-parser.ts
bakkot: This function is only called internally, so it doesn't occur in practice. This is an error you'd only run into if as a developer of ecmarkup, changing the arguments to ecmarkup's CLI.

michaelficarra (approved):

PR #378, opened 2021-11-27
Bump various dependencies

bakkot:
On top of #377.

Including jsdom, which had been blocked on https://github.com/jsdom/jsdom/issues/2968. ~I finally gave up waiting and just reached into the internals to patch it, which is gross, but what can you do.~ jsdom 19 includes [a fix for that issue](https://github.com/jsdom/jsdom/pull/3298), so it can be updated cleanly.


Ms2ger:
jsdom should be better in v19, apparently

bakkot:
Yup, I've included it and dropped the hack.

michaelficarra (approved):

PR #380, opened 2021-11-28
Document structured headers

### LINKED ISSUE #379: Document structured headers
<https://tc39.es/ecmarkup/> should include documentation on structured headers.

gibson042:
Fixes #379

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -177,7 +177,7 @@ toc: false
 
     <emu-clause id="emu-clause-legacy-attributes" legacy>
       <h1>Legacy Attributes</h1>
-      <p><b>aoid:</b> Abstract operation ID. A unique name identifying this clause as an abstract operation. Algorithm steps will auto-link calls to this abstract operation to this clause. If left blank, the aoid will be set to the id of this clause.</p>
+      <p><b>aoid:</b> Abstract operation ID. A unique name identifying this clause as an abstract operation. Algorithm steps will auto-link calls to this abstract operation to this clause. If left blank, the aoid will be set to the id of this clause. When a structured header is present the aoid will be inferred, but this attribute may be set explicitly when not using a structured header.</p>
     </emu-clause>
 
     <emu-clause id="emu-clause-structured-headers">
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -161,20 +161,69 @@ <h1>emu-clause</h1>
 
     <h2>Attributes</h2>
     <p><b>id:</b> Clause id. Must be unique.</p>
-    <p><b>aoid:</b> Abstract operation ID. A unique name identifying this clause as an abstract operation. Algorithm steps will auto-link calls to this abstract operation to this clause. If left blank, the aoid will be set to the id of this clause.</p>
+    <p><b>example:</b> If present, the clause is an example.</p>
+    <p><b>legacy:</b> If present, the clause is Legacy.</p>
     <p><b>normative-optional:</b> If present, the clause is Normative Optional.</p>
+    <p><b>type:</b> Optional: Type of feature described by the clause.</p>
+    <ul>
```
gibson042, line 168: `<p>` elements cannot contain `<ul>` elements, but the structural connection could be captured by replacing this sequence of `<p>`s with a `<dl>`.

spec/index.html
```diff
@@ -161,20 +161,69 @@ <h1>emu-clause</h1>
...
+      <li>"host-defined abstract operation"</li>
+      <li>"implementation-defined abstract operation"</li>
+      <li>"internal method"</li>
+      <li>"numeric method"</li>
+      <li>"sdo" or "syntax-directed operation"</li>
+    </ul>
+
+    <emu-clause id="emu-clause-legacy-attributes" legacy>
+      <h1>Legacy Attributes</h1>
+      <p><b>aoid:</b> Abstract operation ID. A unique name identifying this clause as an abstract operation. Algorithm steps will auto-link calls to this abstract operation to this clause. If left blank, the aoid will be set to the id of this clause.</p>
```
gibson042, line 180: How and where (if at all) should it be stated that structured headers are the replacement for aoid?

bakkot:
```suggestion
      <p><b>aoid:</b> Abstract operation ID. A unique name identifying this clause as an abstract operation. Algorithm steps will auto-link calls to this abstract operation to this clause. If left blank, the aoid will be set to the id of this clause. When a structured header is present the aoid will be inferred, but this attribute may be set explicitly when not using a structured header.</p>
```

bakkot (approved):
Thanks for doing this! I kept meaning to but never quite got around to it.

ExE-Boss:
Note that the “[Structured Headers](https://tc39.es/ecmarkup/#emu-clause-structured-headers)” section applies to `<emu‑annex>` as well.

PR #383, opened 2021-12-09
Fix propagation of user-code across Abstract Closure boundaries

syg:
Was accidentally blocking propagation of all effects.


bakkot (approved):

bakkot:
LGTM but I'm not going to bother upstreaming this right away, since there are no other effects in use.

PR #385, opened 2021-12-10
Fixes to user code effect handling

syg:
Please see individual commit messages.

One notable consequence of 236ce06 is that the `<dt>effects</dt><dd>user-code</dd>` header is redundant in most cases. It is now enough (and easier) to add an `<emu-meta effects="user-code">cause</emu-meta>` around a root cause step, propagation will take care of the rest.

I'm not removing support for it in the structured header parser in this PR. We can do so at a later time if it's not useful to keep around.


bakkot:
> We can do so at a later time if it's not useful to keep around.

Some operations are defined without algorithms; it'll still be useful for those.

bakkot (approved):

PR #388, opened 2021-12-10
Add [fence-effects=""] to stop effect propagation

syg:
This allows for manual fencing where needed, like in GeneratorStart.

### DIFF
--- a/src/Xref.ts
+++ b/src/Xref.ts
@@ -91,18 +91,31 @@ export default class Xref extends Builder {
   shouldPropagateEffect(effectName: string) {
     if (!this.isInvocation) return false;
     if (this.clause) {
-      // Xrefs nested inside Abstract Closures should not propagate the
-      // user-code effect, since those are effectively nested functions.
+      // Xrefs should not propagate past explicit fences in parent steps. Fences
+      // must be at the beginning of steps.
+      //
+      // Abstract Closures are considered automatic fences for the user-code
+      // effect, since those are effectively nested functions.
       //
       // Calls to Abstract Closures that can call user code must be explicitly
       // marked as such with <emu-meta effects="user-code">...</emu-meta>.
-      if (effectName === 'user-code') {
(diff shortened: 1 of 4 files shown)


bakkot:
These are intended to be used only at the beginning of steps, yeah? We already have a different mechanism for metadata about steps, namely `id`s (which look like (`1. [id="foo"] Step.`); I'd prefer to reuse that (as in `1. [fences-effects="user-code"] Step.`).

This would require a change to ecmarkdown; I can make that change if this approach sounds good to you.

syg:
> These are intended to be used only at the beginning of steps, yeah? We already have a different mechanism for metadata about steps, namely ids (which look like (1. [id="foo"] Step.); I'd prefer to reuse that (as in 1. [fences-effects="user-code"] Step.).

Exactly right.

> This would require a change to ecmarkdown; I can make that change if this approach sounds good to you.

I admit the only reason I did it this way is because I didn't want to touch ecmarkdown. If you're offering, doing it in ecmarkdown would be ideal.

bakkot:
Rebased on https://github.com/tc39/ecmarkup/pull/392 and switched to the syntax suggested above.

src/Xref.ts
```diff
@@ -91,21 +91,34 @@ export default class Xref extends Builder {
...
+        ) {
+          return false;
+        }
+
+        if (
+          parent
+            .getAttribute('fence-effects')
+            ?.split(',')
+            .map(s => s.trim())
+            .includes(effectName)
```
syg, line 119: I feel like we don't need structured processing here, and a `parent.getAttribute('fence-effects')?.includes()` suffices. WDYT?

bakkot: Ehh.. that probably does suffice in practice, but it's not like the split + trim is expensive.

syg: Fair enough, also not that many people build ecma262 anyway.

bakkot (approved):

PR #392, opened 2021-12-14
bump ecmarkdown and support arbitrary attributes for steps

bakkot:
For use in #388.

### DIFF
--- a/src/Algorithm.ts
+++ b/src/Algorithm.ts
@@ -11,7 +11,7 @@ function findLabeledSteps(root: EcmarkdownNode) {
   const steps: OrderedListItemNode[] = [];
   emd.visit(root, {
     enter(node: EcmarkdownNode) {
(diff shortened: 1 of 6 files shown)


src/Algorithm.ts
```diff
@@ -80,7 +80,7 @@ export default class Algorithm extends Builder {
       const labeledSteps = findLabeledSteps(emdTree);
       for (const step of labeledSteps) {
         const itemSource = innerHTML.slice(step.location.start.offset, step.location.end.offset);
-        const offset = itemSource.match(/^\s*\d+\. \[id="/)![0].length;
+        const offset = itemSource.match(/^\s*\d+\. \[ *id *= *"/)![0].length;
```
michaelficarra, line 83: This requires `id` to be used as the first attribute?

michaelficarra (approved):
I don't love the regexps, but they're good enough for error reporting.

PR #395, opened 2021-12-16
Ensure emu-meta tags are fenced by the same mechanisms that xrefs are

bakkot:
Previously emu-meta effects propagated through fences.

Note that the first commit is just moving when stuff happens, and has no effect on its own. It should probably be reviewed separately from the third commit, which has the actual changes.


syg (approved):
lgtm! Thank you for fixing.

PR #400, opened 2022-01-06
Add shortcuts help dialog

syg:
Screenshot from 2022-01-05 16-58-44


syg:
@bakkot The multipage baseline is failing in a way I don't quite understand. Maybe this shortcuts `div` needs to be inserted elsewhere?

bakkot:
Running `npm run update-baselines` should fix baselines.

syg:
> Running npm run update-baselines should fix baselines.

Sorry, I meant update-baselines on multipage is failing in a way I don't understand:

```

  1) baselines
       test/baselines/sources/multipage.html:
     Error: non-clause children are not yet implemented: DIV
      at Spec.<anonymous> (lib/Spec.js:615:31)
      at Generator.next (<anonymous>)
      at /usr/local/google/home/syg/tc39/ecmarkup/lib/Spec.js:8:71
      at new Promise (<anonymous>)
      at __awaiter (lib/Spec.js:4:12)
      at Spec.buildMultipage (lib/Spec.js:559:16)
      at Spec.<anonymous> (lib/Spec.js:385:28)
      at Generator.next (<anonymous>)
      at fulfilled (lib/Spec.js:5:58)
```

src/Spec.ts
```diff
@@ -1137,6 +1137,11 @@ ${await utils.readFile(path.join(__dirname, '../js/multipage.js'))}
     const location = this.opts.location;
     const stage = this.opts.stage;
 
+    const shortcutsHelp = this.doc.createElement('div');
+    shortcutsHelp.setAttribute('id', 'shortcuts-help');
+    shortcutsHelp.innerHTML = getBoilerplate('shortcuts-help');
```
bakkot, line 1142: Ideally we'd only mention `m` when `this.opts.multipage` is true (it doesn't do anything otherwise). So probably just inline the boilerplate into a template?

```suggestion
    shortcutsHelp.innerHTML = `
<ul>
  <li><span>Toggle shortcuts help</span><code>?</code></li>
  <li><span>Toggle "can call user code" annotations</span><code>u</code></li>
${this.opts.multipage ? `<li><span>Navigate to/from multipage</span><code>m</code></li>` : ''}
  <li><span>Jump to search box</span><code>/</code></li>
</ul>`;
```

syg: Done.

bakkot:
> Sorry, I meant update-baselines on multipage is failing in a way I don't understand:

Ah. It's complaining because it sees the `div` as being basically a clause, since it's at the top level and comes after the intro, and it doesn't know how to split that into a page.

Presumably we want this to be on every page of the multipage build, like the table of contents and unlike the rest of the boilerplate. So, easiest fix is to move this out of `buildBoilerplate` and into its own step of `build`, and then to add it to the `tocEles` list passed to `buildMultipage` so that it gets cloned onto every page like the table of contents does (which should at that point get renamed to something like `sharedEles`).

syg:
> Presumably we want this to be on every page of the multipage build, like the table of contents and unlike the rest of the boilerplate. So, easiest fix is to move this out of buildBoilerplate and into its own step of build, and then to add it to the tocEles list passed to buildMultipage so that it gets cloned onto every page like the table of contents does (which should at that point get renamed to something like sharedEles).

Perfect, thanks for the help.

bakkot:
Ah jeeze, now it's failing because of a baseline-like test that isn't a baseline. I can push up a fix for that one.

bakkot (approved):
Nice, thanks!

Issue #252, opened 2020-09-07
Biblio file gives precedence to Annex-B productions


bakkot:
I'm told, though I haven't confirmed this, that a proposal which references a production such as `ObjectLiteral` which is redefined in Annex B, the proposal will link to the Annex B version instead of the main spec.

I suspect this is because the biblio file lists linkable parts of the spec in order, and annex B gives new definitions to some parts of the grammar, which causes a new entry for the LHS in the biblio file with the same key as the original.

bakkot:
Fixed by https://github.com/tc39/ecmarkup/pull/405.

Issue #251, opened 2020-09-07
Automatically update ecma262biblio.json


bakkot:
This file is meant to allow proposals which refer to parts of ECMA262 (say, a grammar production) to automatically link to the relevant part of 262, just as if they'd been included in the spec itself. But it's very stale. We should publish the file as its own package and have a script on https://github.com/tc39/ecma262 which updates and publishes a new minor version of that package whenever a commit is added to the main branch of that repository, then add it as either a dependency of this package or a dev-dependency of the proposal template.

ljharb:
Why does it need to be published as opposed to just referenced from the ecma262 repo?

bakkot:
As in, have proposals point to the github repo? I suppose that would also work, but generally I prefer to use the npm registry to manage dependencies, since that is the usual way of doing things.

ljharb:
that'll be a lot of overhead in the ecma262 branch tho, unless we can add CI to require an updated biblio to be in each PR, and even then we'd want a version tag for every bump of that package, which would add a lot of commits to ecma262 just for that purpose.

bakkot:
We don't need any of that. The biblio file doesn't need to be committed and doesn't need to live in any repository. We can just have a Travis job or whatever which regenerates the biblio file and its package.json and publishes it whenever ecma262 master is pushed.

ljharb:
I suppose that would address my concerns, but it seems problematic to me to be publishing a package that has no git sha associated with it.

bakkot:
We can put the ecma262 sha in a README.md in the package.

ljharb:
true, or even populate the package's "repository" link with a direct link to the sha, or something.

bakkot:
It's common practice to have your repository contain source files and have your published package contain build artifacts, e.g. the output of babel or whatever. This doesn't seem particularly different.

rbuckton:
I created https://github.com/tc39/ecma262/issues/2321 in the ecma262 repository for this, as ecmarkup isn't the only tool that could benefit from being able to fetch an up-to-date `ecma262biblio.json` file.

rbuckton:
I've been thinking about dusting off my VSCode plugin for ecmarkup (I already started tinkering with it again a few months ago). One of the things I'd love to be able to do is have the ability to "Insert a spec section" into the current document. That would essentially give you a list of searchable section names in the biblio, then fetch the source for the section and insert it at the cursor location so that you can start editing.

bakkot:
[...]

PR #405, opened 2022-01-24
some tweaks to the biblio

bakkot:
Some notable changes:
- breaking: drops the built-in (but very stale) ecma262 biblio, and consequently also the `--no-ecma-262-biblio option`
- breaking: replaces `--biblio` with `--write-biblio`, which is clearer
- adds `--load-biblio` which takes either a path or an npm package to specify an external biblio, in preparation for https://github.com/tc39/ecmarkup/issues/251 (may be specified multiple times)

Also:
- some tweaks to which things go into the web-shipped version (namely, only the things it uses)
- some tweaks to which things which go into the `--write-biblio` version (again, only the things it uses)
- some refactoring to the biblio structure, most notably getting rid of `extends Array`; also renames the namespace holding entries from external biblios from "global" to "external".


test/baselines/generated-reference/autolinking.html
```diff
@@ -8,7 +8,7 @@
 </ul></div><div id="spec-container">
 <emu-clause id="sec-foo">
   <h1><span class="secnum">1</span> Autolinking</h1>
-  <p>Type, type, <emu-xref aoid="Type" id="_ref_0"><a href="https://tc39.es/ecma262/#sec-ecmascript-data-types-and-values">Type</a></emu-xref>(), type()</p>
+  <p>Type, type, <emu-xref aoid="Type"><a href="https://tc39.es/ecma262/#sec-ecmascript-data-types-and-values">Type</a></emu-xref>(), type()</p>
```
bakkot, line 11: This change was due to the "move logic for registering op biblio entries" commit. 

Background: For `xref`s to terms defined in the current document, we add an `id` to the `xref` so we can link to it from the References pane when hovering the definition of the term. That's what the `id="_ref_0"` attributes are.

Prior to the referenced commit, we created the entry for an AO as part of adding the entry for a clause; that meant that when loading an external biblio containing an entry for a clause, we'd re-recreate the entry for its AO (thus shadowing the original one, from the external biblio). But all created newly-entries are considered to be in the _current_ namespace, so the analysis thought the AO was defined in the current document, and hence `xref`s to it should be given `id`s. 

The tweak in that commit means we don't re-create the biblio entry for the Type AO, and just use the original, external one. As a consequence, the analysis correctly recognizes that this xref is to an external term and so does not need to be given an `id`. Hence the diff.

bakkot:
Also I think this fixes #252 (the "only include items in top-level namespace" commit).

src/cli.ts
```diff
@@ -107,6 +107,17 @@ const build = debounce(async function build() {
     }
     let warned = false;
 
+    for (let toResolve of args['load-biblio'] ?? []) {
+      if (/^[./]/.test(toResolve)) {
```
michaelficarra, line 111:
```suggestion
      if (/^\.{0,2}\//.test(toResolve)) {
```

Maybe also pull the regexp out since it's used elsewhere.

michaelficarra (approved):

PR #408, opened 2022-02-19
completion-record local typechecking

bakkot:


src/Clause.ts
```diff
@@ -316,3 +324,17 @@ export default class Clause extends Builder {
...
+      name: p.name,
+      type: p.type,
+    })),
+    optionalParameters: parsedHeader.optionalParams.map(p => ({
+      name: p.name,
+      type: p.type,
+    })),
+    return: parsedHeader.returnType,
+  };
+}
```
arhadthedev: An empty line is forgotten here. BTW, maybe a check should be added into a linting stage?

src/Biblio.ts
```diff
@@ -280,9 +282,49 @@ export interface BiblioEntryBase {
...
+      typeOfValueIfNormal: Type | null;
+      completionType: 'normal' | 'mixed';
+    }
+  | {
+      kind: 'list';
+      elements: Type | null;
+    }
+  | {
+      kind: 'union';
+      types: Exclude<Type, { kind: 'union' }>[];
```
michaelficarra, line 308: Can we exclude `{ kind: 'completion' }` here too?

src/Clause.ts
```diff
@@ -300,11 +342,26 @@ export default class Clause extends Builder {
...
+        if (
+          signature?.return?.kind === 'union' &&
+          signature.return.types.some(e => e.kind === 'completion') &&
+          signature.return.types.some(e => e.kind !== 'completion')
+        ) {
+          spec.warn({
+            type: 'node',
+            node: clause.header!,
+            ruleId: 'completion-union',
+            message: `algorithms should return either completions or things which are not completions, never both`,
```
michaelficarra, line 362: I think this warning would be unnecessary if we can get the type system to not even allow construction of this kind of signature.

src/Spec.ts
```diff
@@ -539,6 +548,240 @@ export default class Spec {
...
+      // checks elsewhere ensure that well-formed documents never have a union of completion and non-completion, so checking the first child suffices
+      // TODO: this is for 'a break completion or a throw completion', which is kind of a silly union; maybe address that in some other way?
+      const isCompletion =
+        returnType.kind === 'completion' ||
+        (returnType.kind === 'union' && returnType.types[0].kind === 'completion');
+      if (['Completion', 'ThrowCompletion', 'NormalCompletion'].includes(xref.aoid)) {
+        if (consumedAsCompletion) {
+          // TODO mf to improve message
+          warn(
+            `${xref.aoid} is explicitly a Completion Record and should not be unwrapped or marked`
```
michaelficarra, line 697:
```suggestion
            `${xref.aoid} clearly creates a Completion Record; it does not need to be marked as such, and it would not be useful to immediately unwrap its result`
```

src/Spec.ts
```diff
@@ -539,6 +548,240 @@ export default class Spec {
...
+          );
+        }
+      } else if (isCompletion && !consumedAsCompletion) {
+        warn(`${xref.aoid} returns a Completion Record, but is not consumed as one`);
```
michaelficarra, line 701:
```suggestion
        warn(`${xref.aoid} returns a Completion Record, but is not consumed as if it does`);
```

src/Spec.ts
```diff
@@ -539,6 +548,240 @@ export default class Spec {
...
+      } else if (!isCompletion && consumedAsCompletion) {
+        warn(`${xref.aoid} does not return a Completion Record, but is consumed as one`);
```
michaelficarra, line 703:
```suggestion
        warn(`${xref.aoid} does not return a Completion Record, but is consumed as if it does`);
```

src/Spec.ts
```diff
@@ -539,6 +548,240 @@ export default class Spec {
...
+      }
+      if (returnType.kind === 'unused' && !isCalledAsPerform(xref, false)) {
+        warn(
+          `${xref.aoid} does not return a value and should only be invoked as \`Perform ${xref.aoid}(...).\``
```
michaelficarra, line 707:
```suggestion
          `${xref.aoid} does not return a meaningful value and should only be invoked as \`Perform ${xref.aoid}(...).\``
```

src/Spec.ts
```diff
@@ -539,6 +548,240 @@ export default class Spec {
...
+      }
+      if (
+        alwaysAssertedToBeNormal.has(xref.aoid) &&
+        alwaysAssertedToBeNormal.get(xref.aoid) !== 'top'
+      ) {
+        const old = alwaysAssertedToBeNormal.get(xref.aoid);
+        const asserted = isAssertedToBeNormal(xref);
+        if (!asserted) {
+          alwaysAssertedToBeNormal.set(xref.aoid, 'top');
+        } else if (old === null) {
```
michaelficarra, line 728: These `old == null` checks aren't necessary for correctness, but they're fine I guess.

src/Spec.ts
```diff
@@ -539,6 +548,240 @@ export default class Spec {
...
+          alwaysAssertedToBeNormal.set(xref.aoid, 'always asserted normal');
+        }
+      }
+    }
+
+    for (const [aoid, state] of onlyPerformed) {
+      if (state !== 'only performed') {
+        continue;
+      }
+      const message = `${aoid} is only ever invoked with Perform, so it should return ~unused~ or a Completion Record normally containing ~unused~`;
```
michaelficarra, line 738:
```suggestion
      const message = `${aoid} is only ever invoked with Perform, so it should return ~unused~ or a normal completion containing ~unused~`;
```

Side note: it should be okay to always use perform with an AO that always returns an abrupt completion. We could tweak `isUnused` above to allow for that, but we probably don't need this message to mention that case.

src/header-parser.ts
```diff
@@ -5,29 +5,41 @@ type ParseError = {
...
-      errors: ParseError[];
+      wrappingTag: 'ins' | 'del' | 'mark' | null;
+      type: null;
     }
+  | {
+      name: string;
+      wrappingTag: 'ins' | 'del' | 'mark' | null;
+      type: string;
+      typeOffset: number;
+    };
```
michaelficarra, line 19: You can restructure this signature to use `&` to deduplicate the shared parts.

src/type-parser.ts
```diff
@@ -0,0 +1,315 @@
...
+      if (!(this.remainder === '' || this.remainder[0] === ')' || this.eat(/^,/) != null)) {
+        throw new ParseError(`expecting a parenthesis after a "but not" clause`, this.offset);
+      }
+    }
+    if (opaqueThing.type === '~unused~') {
+      return {
+        kind: 'unused',
+      };
+    } else if (opaqueThing.type === '') {
+      throw new ParseError(`expected to find at type, got empty string`, opaqueStart);
```
michaelficarra, line 209:
```suggestion
      throw new ParseError(`expected to find a type, got empty string`, opaqueStart);
```

test/type-parser.js
```diff
@@ -0,0 +1,237 @@
...
+          },
+        ],
+      });
+
+      assert.deepStrictEqual(TypeParser.parse(`a List`), {
+        kind: 'list',
+        elements: null,
+      });
+
+      assert.deepStrictEqual(TypeParser.parse(`a List of X`), {
```
michaelficarra, line 166: I'd like to see "a List of Lists of X" or similar. Also, a test that covers "records with fields".

michaelficarra (approved):
LGTM, nothing fatal.

src/Spec.ts
bakkot: I much prefer the `=== null`, since it's the bottom type. Plus it's just confusing to set the type to itself, which omitting the condition or having `old !== 'top'` would do.

src/Spec.ts
bakkot: It's also ok to use it with an AO that returns either a normal completion containing unused _or an abrupt completion_. Changed to `... a Completion Record which, if normal, contains ~unused~`.

test/typecheck.js
```diff
@@ -0,0 +1,510 @@
...
+        </dl>
+        <emu-alg>
+          1. Return Foo().
+        </emu-alg>
+        </emu-clause>
+      `);
+    });
+  });
+
+  describe('AO returning unused is not Performed', () => {
```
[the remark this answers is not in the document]
bakkot: No, if it returns a completion record it's reasonable to capture the result and check its [[Type]] in a subsequent step. Not something we can automate.

src/Biblio.ts
bakkot: It's not totally unreasonable to have a parameter which is either a completion record or null, so I'm not going to do this right now.

src/type-parser.ts
```diff
@@ -0,0 +1,315 @@
...
+      };
+    }
+    if (this.eat(/^(a List|Lists)\b/i) != null) {
+      return {
+        kind: 'list',
+        elements: null,
+      };
+    }
+    {
+      const match = this.eat(/^(a Record|Records) with field(s?)\b/i);
```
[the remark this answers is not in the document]
bakkot: We were supposed to, though! The test for `match[1]` below used to refer to this group. Fixed.

PR #414, opened 2022-03-07
Minor cleanups

ptomato:
I came here to fix the step IDs to be valid, but here are a couple more things I noticed while doing that.


bakkot (approved):

bakkot:
Thanks!

PR #422, opened 2022-03-21
bump chalk dependency

bakkot:
V5 is out, but the difference is that it's esm-only, which is not something we care about. v4 is already in the dependency tree; best to stick with that.


ljharb (changes requested):
yep, def stick with chalk v4 for the duration

PR #423, opened 2022-03-21
fix searching for clause numbers

bakkot:
Without this searching for strings of digits with `.`s brings up results like "10.1.1.1 undefined" or whatever instead of "10.1.1.1 OrdinaryGetPrototypeOf ( O )".

Similar issue to https://github.com/tc39/ecmarkup/pull/370. I didn't notice there was separate logic for searching for clause numbers in particular, sigh.


jlachlan (commented):
Not secure

PR #424, opened 2022-03-23
change rule for comparisons against -0

bakkot:
I'm calling this major because it will break the build if you're using `--strict`, although we're not always consistent about that.

Includes version bump commit, so please **rebase**, not squash.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "ecmarkup",
-	"version": "10.0.2",
+	"version": "11.0.0",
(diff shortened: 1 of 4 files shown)


michaelficarra (approved):

jlachlan (commented):
Undo the changes by bro not with 10x as of Nov 2021 there was a bunch of branches and changes trying to catch up on what was switch and why they logged spelling to go and redo bad things

jlachlan (commented):
You changed the old version and marked it as version 10 to 11
image

PR #426, opened 2022-03-27
Put effects in biblio

bakkot:
Based on https://github.com/tc39/ecmarkup/pull/425.

This makes the effects (right now, just "user code") be present in the exported biblio, for the benefit of other tooling and so that proposals can get the [UC] annotation.

However, because this requires additional work on behalf of consumers for the rendering to be accurate (namely, explicitly marking any roots of user code in their own text), this PR also makes the rendering opt-in. This is therefore a breaking change (because 262 will need to opt in to get its current behavior).

### DIFF
--- a/src/Spec.ts
+++ b/src/Spec.ts
@@ -301,7 +301,7 @@ export default class Spec {
   _prodRefs: ProdRef[];
   _textNodes: { [s: string]: [TextNodeContext] };
   _effectWorklist: Map<string, Clause[]>;
-  _effectfulAOs: Map<string, Clause>;
+  _effectfulAOs: Map<string, string[]>;
   _emuMetasToRender: Set<HTMLElement>;
   _emuMetasToRemove: Set<HTMLElement>;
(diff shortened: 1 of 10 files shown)


src/Spec.ts
```diff
@@ -1321,6 +1322,17 @@ ${this.opts.multipage ? `<li><span>Navigate to/from multipage</span><code>m</cod
     const biblios = biblioContents.map(c => JSON.parse(c) as ExportedBiblio);
     for (const biblio of biblios.concat(this.opts.extraBiblios ?? [])) {
       this.biblio.addExternalBiblio(biblio);
+      for (const entry of biblio.entries) {
+        if (entry.type === 'op' && entry.effects?.length > 0) {
+          this._effectfulAOs.set(entry.aoid, entry.effects);
+          for (const effect of entry.effects) {
+            if (!this._effectWorklist.has(effect)) {
```
michaelficarra, line 1329: We really need a MultiMap...

michaelficarra (approved):

Issue #398, opened 2021-12-16
Operations defined by structured header should support an aoid override


gibson042:
The Temporal proposal has a need to mention both the 262 and 402 flavors of some operations, which ecmarkup rejects as duplicate definitions. This can be addressed for traditionally-defined operations via an aoid suffix, but an attempt to do the same for a structured operation is rejected ("_nodes with structured headers should not include an AOID_"). I'm attempting to work around this in https://github.com/tc39/proposal-temporal/pull/1920/commits/82f542b2765314767577c0ec4a5a65b1326a7afc by effectively suppressing the structured definition, but it would be better if instead something like `<emu-clause id="sup-canonicalizetimezonename" type="abstract operation" aoid="CanonicalizeTimeZoneName402">` were supported.

bakkot:
Can you say more about why you need these to be AOIDs?

gibson042:
They don't need to be AOIDs specifically, that just seemed like an obvious option.

gibson042:
I am trying to have two abstract operations in the same proposal spec which are both defined by structured header including the same name (one for showing changes in ECMA-262, the other for showing changes in ECMA-402) but _do not result in an error from `ecmarkup --lint-spec --strict`_, ideally because ecmarkup has a way to recognize them as distinct despite that identical name. This is not currently possible with structured headers, but _is_ possible without them:
```html
<emu-clause id="sec-temporal-isbuiltincalendar" aoid="IsBuiltinCalendar">
  <h1>IsBuiltinCalendar ( _id_ )</h1>
```
```html
<emu-clause id="sup-temporal-isbuiltincalendar" aoid="IsBuiltinCalendar402">
  <h1>IsBuiltinCalendar ( _id_ )</h1>
```

bakkot:
Ah, ok. And to be clear, in the snippet you give without structured headers, you're not actually using the AOID for anything, right? That is, you don't expect to be able to write `IsBuiltinCalendar402("foo")`  in an algorithm step and have it link to that second definition?

The solution I'm currently envisioning is a field in the `dl` along the lines of `<dt>redefinition</dt><dd>true</dd>`, which if present would suppress that warning.

Such a field could also cause ecmarkup to
- automatically add a note to the original section which links to the other definition(s)
- check that the signatures are compatible

unless you're opposed.

But this solution would not have the effect of allowing you to refer to it and have the reference automatically link, like happens with `IsBuiltinCalendar402` in the snippet you've given.

[...]

PR #428, opened 2022-03-28
add redefinition support to structured headers

bakkot:
Fixes https://github.com/tc39/ecmarkup/issues/398 (cc @gibson042).

In that issue I suggested that marking a node as a redefinition could cause other checks to happen. In the end I decided not to do that, since a redefinition could in principle change the signature. All this does is prevent the `aoid` attribute from being generated (and consequently the error from having duplicate aoids).


gibson042 (approved):
LGTM, thanks!

PR #440, opened 2022-04-24
support for SOURCE_DATE_EPOCH

### LINKED ISSUE #390: Determine draft date by the last modified date of the ecmarkup
Specifically, in a proposal, every time I run `ecmarkup spec.emu index.html`, the HTML updates with the current date, even if the ecmarkup hasn't changed.

Can the date generated be "when the input was modified" and not "the current date", to reduce diff churn?

bakkot:
Fixes #390, cc @ljharb.

This lets you do e.g. `SOURCE_DATE_EPOCH=$(date -r in.emu '+%s') ecmarkup in.emu out.html` to set the date in the rendered specification to the last modified date of `in.emu`, for reproducible builds.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -150,6 +150,14 @@ const build = debounce(async function build() {
       warnings.push(err);
     };
 
+    if (process.env.SOURCE_DATE_EPOCH) {
+      const ts = +process.env.SOURCE_DATE_EPOCH;
+      if (ts !== Math.floor(ts)) {
+        fail(`SOURCE_DATE_EPOCH value ${process.env.SOURCE_DATE_EPOCH} is not valid`);
+      }
+      opts.date = new Date(ts * 1000);
+    }
+
     const spec = await ecmarkup.build(args.files[0], utils.readFile, opts);
 
     if (args.verbose) {
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -151,10 +151,11 @@ const build = debounce(async function build() {
     };
 
     if (process.env.SOURCE_DATE_EPOCH) {
-      const ts = +process.env.SOURCE_DATE_EPOCH;
-      if (ts !== Math.floor(ts)) {
-        fail(`SOURCE_DATE_EPOCH value ${process.env.SOURCE_DATE_EPOCH} is not valid`);
+      const sde = process.env.SOURCE_DATE_EPOCH.trim();
+      if (!/^[0-9]+/.test(sde)) {
+        fail(`SOURCE_DATE_EPOCH value ${sde} is not valid`);
       }
+      const ts = +sde;
       opts.date = new Date(ts * 1000);
     }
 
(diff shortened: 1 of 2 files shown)


src/cli.ts
```diff
@@ -150,6 +150,14 @@ const build = debounce(async function build() {
       warnings.push(err);
     };
 
+    if (process.env.SOURCE_DATE_EPOCH) {
+      const ts = +process.env.SOURCE_DATE_EPOCH;
+      if (ts !== Math.floor(ts)) {
+        fail(`SOURCE_DATE_EPOCH value ${process.env.SOURCE_DATE_EPOCH} is not valid`);
+      }
```
ljharb, line 157: Is this meant to allow the empty string, and "Infinity", and `1e3`, and `0xF`?

bakkot: It doesn't allow the empty string (it won't pass the check above), but I suppose I may as well guard against the others. Updated.

ljharb: `+'' !== Math.floor(+'')` is false

src/cli.ts
```diff
@@ -150,6 +150,15 @@ const build = debounce(async function build() {
       warnings.push(err);
     };
 
+    if (process.env.SOURCE_DATE_EPOCH) {
+      const sde = process.env.SOURCE_DATE_EPOCH.trim();
+      if (!/^[0-9]+/.test(sde)) {
```
ljharb, line 155:
```suggestion
      if (!/^[+-]?[0-9]+/.test(sde)) {
```
i suppose :-)

ljharb (approved):

src/cli.ts
bakkot: Yeah but so is `if ('')`.

src/cli.ts
bakkot: No, there's an [actual specification and everything](https://reproducible-builds.org/specs/source-date-epoch/), which says the format must be "identical to the output format of `date +%s`", which doesn't include a leading +/-.

ljharb:
@bakkot this helps, but it'd be much easier to make it work on windows as well if it was something ecmarkup could do - either by default, or via a flag.

bakkot:
@ljharb I am pretty sure you can set environment variables on Windows; what makes Windows special here?

ljharb:
`date` isn't available on windows, and i think the syntax to set env vars on windows is different, but maybe cross-env can help.

There's also the issue i'm seeing in the pattern matching repo where despite builds now being reproducible on my machine, CI still finds a difference.

bakkot:
I am pretty sure there's a way to get the equivalent on Windows, though I don't have a machine in front of me to test. I would very much prefer to stick to the standard option here, if it's at all practical, rather than adding more CLI options.

If you're successfully setting the relevant environment variable and still seeing builds not being reproducible I can look into that; open an issue with a repro?

ljharb:
Filed #470.

PR #445, opened 2022-05-06
Stop rejecting expanded-year dates as positive-signed real numbers

### LINKED ISSUE #442: False positive with "real numbers must not start with +" lint rule
The linter rejects ISO 8601 notation like +275760-09-13T00:00:00Z as being a positive real number with a leading plus sign.

gibson042:
Fix #442

### DIFF
--- a/src/lint/collect-spelling-diagnostics.ts
+++ b/src/lint/collect-spelling-diagnostics.ts
@@ -42,7 +42,8 @@ const matchers = [
   },
   {
     // this needs its own rule to catch +0 as a real number
-    pattern: /(?<= )\+[0-9]/gu,
+    // (but not similar text such as expanded-year dates like +000000-01-01)
+    pattern: /(?<= )\+[0-9](?![0-9]*[+-])/gu,
     message: 'positive real numbers should not have a leading plus sign (+)',
   },
   {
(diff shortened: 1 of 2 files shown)


bakkot (approved):

src/lint/collect-spelling-diagnostics.ts
```diff
@@ -42,7 +42,8 @@ const matchers = [
   },
   {
     // this needs its own rule to catch +0 as a real number
-    pattern: /(?<= )\+[0-9]/gu,
+    // (but not similar text such as expanded-year dates like +000000-01-01)
+    pattern: /(?<= )\+[0-9](?![0-9]*[+-])/gu,
```
michaelficarra, line 46: What is the `+` for here?

```suggestion
    pattern: /(?<= )\+[0-9](?![0-9]*-)/gu,
```

gibson042: Hmm, I think I intended for that to be a `:` to prevent this rule from rejecting isolated UTC offsets like "+00:00".

gibson042: #463

PR #452, opened 2022-05-24
Link "min" and "max" even when not followed by parentheses

gibson042:
Unlike "Call" and "Set" and "UTC", they are not used in either ECMA-262 or ECMA-402 except in reference to the specific corresponding operation, and it's weird that they aren't linked at the end of [Mathematical Operations](https://tc39.es/ecma262/multipage/notational-conventions.html#sec-mathematical-operations) in the way that "abs" and "floor" are.


bakkot (approved):
Fine by me.

PR #457, opened 2022-06-10
Document the "effects" structured header key

gibson042:
Ref 4f14b656fc9882eff7fc5c96bfc54f147148bb2f


bakkot (approved):
This is mentioned under [Effects](https://tc39.es/ecmarkup/#effects), but yeah, should be here too.

PR #464, opened 2022-07-05
Expression parser

bakkot:
This adds a barebones parser for algorithm steps, recognizing calls, SDO invocations, records, and lists. It is invoked downstream of the biblio being built, which means it can differentiate `List of _bar_` (which is not an SDO invocation) from `SDO of _bar_` (which is).

The parser is much less sophisticated than the ones in ecmaspeak-py or esmeta. On the other hand, it's also much less specialized to the current spec text, so hopefully will not require updates as often.

The new parser is used to
- ensure parentheses are balanced and so on
- check that everything being called is defined (with a hardcoded list of exceptions, see below)
- check that AOs and SDOs are invoked with the appropriate number of arguments
- check that record field names are not duplicated within a record
- fix the [super gross](https://github.com/tc39/ecmarkup/blob/e83854e4c9597a1e1fec96f258645481325713ce/src/Spec.ts#L630-L635) post-render walking we were previously doing to check that AOs returning completion records were invoked appropriately, which also lets us get more precise location information for those errors

I've checked this on 262 and 402 (which pass without error) as well as proposal-temporal (which has a few genuine bugs this catches).

It also incidentally adds AO kinds ("syntax-directed operation", "host-defined abstract operation", etc) to the biblio. This is not a breaking change because the new type allows the kind to be `undefined`.

---
Unfortunately I've had to hardcode the names of certain operations which are invoked without being defined as operations, namely
 - thisTimeValue
 - thisStringValue
 - thisBigIntValue
 - thisNumberValue
 - thisSymbolValue
 - thisBooleanValue
 - toUppercase
 - toLowercase
 - ℝ
 - 𝔽
 - ℤ

The `thisXValue` ones are a [weird thing we do](https://tc39.es/ecma262/multipage/numbers-and-dates.html#thistimevalue) where we define an AO without giving it its own clause, represented by source text like

```
The abstract operation <dfn id="thistimevalue" aoid="thisTimeValue" oldids="sec-thistimevalue">thisTimeValue</dfn> takes argument _value_. It performs the following steps when called:
```

We should get rid of these. (Also note that `aoid` on `dfn` doesn't do anything.)

`toUppercase` and `toLowercase` are invoking Unicode algorithms (see e.g. step 5 of [Canonicalize](https://tc39.es/ecma262/multipage/text-processing.html#sec-runtime-semantics-canonicalize-ch)). Not sure what to do about those - maybe those names should be written as `` `code` ``?

`ℝ` and friends just need to be defined with `emu-eqn` and then they can be removed from the hardcoded list; I'll make that change once https://github.com/tc39/ecma262/pull/2810 lands.

---
Depends on https://github.com/tc39/ecmarkdown/pull/92.

### DIFF
--- a/src/Spec.ts
+++ b/src/Spec.ts
@@ -45,16 +45,11 @@ import {
 import { lint } from './lint/lint';
 import { CancellationToken } from 'prex';
 import type { JSDOM } from 'jsdom';
-import type {
-  OrderedListItemNode,
-  OrderedListNode,
-  parseAlgorithm,
-  UnorderedListItemNode,
-} from 'ecmarkdown';
+import type { OrderedListNode, parseAlgorithm } from 'ecmarkdown';
 import { getProductions, rhsMatches, getLocationInGrammarFile } from './lint/utils';
 import type { AugmentedGrammarEle } from './Grammar';
 import { offsetToLineAndColumn } from './utils';
-import { parse as parseExpr, walk as walkExpr, Expr } from './expr-parser';
+import { parse as parseExpr, walk as walkExpr, Expr, PathItem } from './expr-parser';
 
 const DRAFT_DATE_FORMAT: Intl.DateTimeFormatOptions = {
   year: 'numeric',
@@ -484,8 +479,7 @@ export default class Spec {
 
     if (this.opts.lintSpec) {
       this.log('Checking types...');
-      this.checkCompletionUsage();
-      this.checkCalls();
+      this.typecheck();
     }
     this.log('Propagating effect annotations...');
     this.propagateEffects();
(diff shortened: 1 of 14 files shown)


src/Spec.ts
```diff
@@ -598,146 +599,174 @@ export default class Spec {
...
-            // @ts-ignore
-            path.unshift([].indexOf.call(pointer.parentElement!.children, pointer));
-          }
-          if (pointer.tagName === 'EMU-ALG') {
-            alg = pointer;
-            break;
-          }
-          pointer = pointer.parentElement;
+      const expressionVisitor = (expr: Expr, path: PathItem[]) => {
+        // console.log(require('util').inspect(expr, { depth: Infinity }));
```
michaelficarra, line 623:
```suggestion
```

src/Spec.ts
```diff
@@ -598,146 +599,174 @@ export default class Spec {
...
+            this.warn({
+              type: 'contents',
+              ruleId: 'expression-parsing',
+              message: item.message,
+              node,
+              nodeRelativeLine: line,
+              nodeRelativeColumn: column,
+            });
+          } else {
+            // console.log(require('util').inspect(item, { depth: Infinity}));
```
michaelficarra, line 761:
```suggestion
```

michaelficarra (approved):

PR #471, opened 2022-07-13
fix emu-import bug

acutmore:
Today I was trying to update https://github.com/tc39/proposal-record-tuple from `ecmarkup@9.1.0` to the latest but from version `10.0.1` and onwards the generated spec was almost entirely empty as all the `<emu-import>` tags were being ignored.

Stepping through with a debugger I get to the `querySelectorAll("EMU-IMPORT")` and see that it returns an empty NodeList. Changing it to lowercase fixes the issue, I'm presuming there is some recent non-locked down sub dependency of jsdom that has recently changed, otherwise I don't know how this is working for others.

screenshot-1

screenshot-2

### DIFF
--- a/src/Spec.ts
+++ b/src/Spec.ts
@@ -1931,7 +1931,7 @@ function getBoilerplate(file: string) {
 }
 
 async function loadImports(spec: Spec, rootElement: HTMLElement, rootPath: string) {
-  const imports = rootElement.querySelectorAll('EMU-IMPORT');
+  const imports = rootElement.querySelectorAll('emu-import');
   for (let i = 0; i < imports.length; i++) {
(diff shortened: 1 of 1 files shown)


bakkot:
Sorry, yeah, this is a [bug in a transitive dependency](https://github.com/dperini/nwsapi/issues/57), which updated fairly recently. It's working for others because they still have a package-lock which names the older version of nwsapi, presumably. I'd prefer to just pin version 2.2.0 of nwsapi until this bug is fixed, rather than changing the selector. (You can do that in your own package.json right now without needing ecmarkup to make that change.)

acutmore:
Cool. I'll do that instead. Thanks!

PR #473, opened 2022-07-24
Allow specifying custom clause numbers

nicolo-ribaudo:
Some proposal authors prefer writing proposals using the same section numbers as ecma262, but it requires [hacks](https://github.com/tc39/proposal-symbol-predicates/blob/372c0528da010c15efcee449fce6045825d9ec80/spec.emu#L13-L44).

This PR allows specifying a custom `number="..."` attribute for `emu-clause` nodes.
- for nested clauses, you can only specify the last number segment: the first ones are always inherited from the parent clause
- after a custom number, automatic numbering continues from that number and not from where it left. This means that `<emu-clause /> <emu-clause number="8" /> <emu-clause />` gives `1, 8, 9` and not `1, 8, 2` or `1, 8, 3`.

cc @ljharb

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -171,6 +171,7 @@ toc: false
     <p><b>example:</b> If present, the clause is an example.</p>
     <p><b>legacy:</b> If present, the clause is Legacy.</p>
     <p><b>normative-optional:</b> If present, the clause is Normative Optional.</p>
+    <p><b>number:</b> Optional: An explicit clause number, overriding the default auto-incrementing number. Can be a nested number, as in `number="2.1"`.</p>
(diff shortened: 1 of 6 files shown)


ljharb:
Yay!

bakkot (approved):

bakkot:
Thanks! I pushed a commit adding support for nested numbers, as in `number="1.1"`, as well as a few more warnings for error cases.

bakkot:
Released in 14.1.0.

nicolo-ribaudo:
Thank you!

ljharb:
This is great, and works perfectly! I’ve cleaned up 3 proposals so far.

PR #474, opened 2022-07-25
CSS: Fix very long words being cut off (word-wrap: break-word)

12Me21:
By default, CSS won't insert linebreaks in words that are too long to fit, resulting in text overflowing or being cut off.
For example: in [ECMAScript, B.2.1.1 - escape](https://tc39.es/ecma262/multipage/additional-ecmascript-features-for-web-browsers.html#sec-escape-string), the end of this string is lost offscreen, if your screen isn't wide enough:
> image

[`word-wrap`](https://drafts.csswg.org/css-text/#overflow-wrap-property) only takes effect when a word would otherwise overflow, and only when `white-space` allows text wrapping. So, it should be safe to apply to the entire document.

### DIFF
--- a/css/elements.css
+++ b/css/elements.css
@@ -1,5 +1,6 @@
 body {
   display: flex;
+  word-wrap: break-word;
   font-size: 18px;
   line-height: 1.5;
   font-family: Cambria, Palatino Linotype, Palatino, Liberation Serif, serif;


gibson042 (approved):

bakkot (approved):

bakkot:
Thanks!

Issue #476, opened 2022-08-17
validate HTML entities


michaelficarra:
When I accidentally write `&lte;` when I mean `&le;`, ecmarkup should complain.

bakkot:
`&lte;` is a particularly tricky one because - fun fact - `&lte;` decodes to `<e;`. [Certain entities](https://html.spec.whatwg.org/multipage/named-characters.html#named-character-references) do not require a trailing semicolon, though [this is](https://html.spec.whatwg.org/multipage/parsing.html#named-character-reference-state) a "missing-semicolon-after-character-reference parse error".

Should be possible to differentiate, but the [package all the parsers use](https://www.npmjs.com/package/entities) doesn't, so it might require a bit of work.

PR #481, opened 2022-09-04
formatter: render HTML entities

### LINKED ISSUE #477: formatter should translate HTML entities to code points where possible
cf https://github.com/tc39/ecma262/issues/404#issuecomment-1218537924

See also https://github.com/tc39/ecmarkup/issues/476.

Note to self: this should not include blank codepoints like `&nbsp;` or `&lrm;`. "blank" [probably](http://www.unicode.org/faq/unsup_char.html) means "[Control](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=[:General_Category=Control:]), [White_Space](https://util.unicode.org/UnicodeJsps/properties.jsp?a=White_Space), or [Default_Ignorable](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=[:Identifier_Type=Default_Ignorable:])", in Unicode terms. (Linked page does not include "Control" but it seems an obvious thing to add; it is not a subset of DI, per [its definition](https://unicode.org/reports/tr44/#Default_Ignorable_Code_Point), and indeed `\u0000` is not DI.)

Also there's 93 entities which expand to multiple code points, often including a variation selector or "U+0338 COMBINING LONG SOLIDUS OVERLAY". Should exclude any multi-code point ligatures where at least one code point is blank in the above sense, and probably any which include combining characters (i.e. `gc=M`).

I.e. exclude anything which expands to a sequence of code points which matches `/\p{White_Space}|\p{DI}|\p{gc=M}|\p{gc=C}/u`.

bakkot:
Fixes #477.


filter-entities.js
```diff
@@ -0,0 +1,19 @@
...
+// entities.json is from https://html.spec.whatwg.org/entities.json
+// which is built by https://github.com/whatwg/html-build/tree/main/entities
+// this processes it into a form more useful for us
+// no new entities will be added, so the output of this file is committed
+// rather than running it every time
+
+let fs = require('fs');
+let entities = require('./entities.json');
+
+let transformed = Object.fromEntries(Object.entries(entities).map(([k, v]) => {
```
michaelficarra, line 12: You could `flatMap` here to omit them instead of mapping them to `null`.

bakkot: I wanted to keep the keys present so I could use the same file to do https://github.com/tc39/ecmarkup/issues/476.

michaelficarra (approved):

Issue #129, opened 2017-11-29
warn when variables appear only once


michaelficarra:
I can't imagine a use case for a variable which appears only once, and this would have caught a recent typo in one of my proposals.

ljharb:
Do you mean once including the declaration?

I often use a variable only once, but certainly if a variable is declared but unused, a warning would be great.

PR #483, opened 2022-09-05
lint for used-but-not-declared and declared-but-not-used variables in algorithms

bakkot:
Based on https://github.com/tc39/ecmarkup/pull/482. Fixes https://github.com/tc39/ecmarkup/issues/129.

Given https://github.com/tc39/ecma262/pull/2896 and https://github.com/tc39/ecma262/commit/5f759fa947f6c8ea2bf29c7f5b3a345a8f7714db, this passes on ecma262 except for a few genuinely unused declarations, which I'll fix as part of merging this.

This is pretty conservative (e.g. it assumes any variable referred to in a previous paragraph in the clause is defined, to handle stuff like [this](https://tc39.es/ecma262/multipage/memory-model.html#sec-tear-free-aligned-reads)), but seems to work well. In particular it handles abstract closures correctly (variables must be captured from the outer scope, captures count as using variables from the outer scope).

### DIFF
--- a/src/lint/rules/variable-use-def.ts
+++ b/src/lint/rules/variable-use-def.ts
@@ -1,0 +1,341 @@
+import type { FragmentNode, OrderedListNode, UnorderedListNode, TextNode, UnderscoreNode } from 'ecmarkdown';
+import type { Reporter } from '../algorithm-error-reporter-type';
+
+export function checkVariableUsage(
+  containingAlgorithm: Element,
+  steps: OrderedListNode,
+  report: Reporter
+) {
+  if (containingAlgorithm.hasAttribute('replaces-step')) {
+    // TODO someday lint these by doing the rewrite (conceptually)
+    return;
+  }
+  let scope = new Scope(report);
+
+  let parentClause = containingAlgorithm.parentElement;
+  while (parentClause != null) {
+    if (parentClause.nodeName === 'EMU-CLAUSE') {
+      break;
+    }
+    if (parentClause.nodeName === 'EMU-ANNEX') {
+      // Annex B adds algorithms in a way which makes it hard to track the original
+      // TODO someday lint Annex B
+      return;
+    }
+    parentClause = parentClause.parentElement;
+  }
+  // we assume any name introduced earlier in the clause is fair game
+  // this is a little permissive, but it's hard to find a precise rule, and that's better than being too restrictive
+  let preceding = previousOrParent(containingAlgorithm, parentClause);
+  while (preceding != null) {
+    if (preceding.tagName !== 'EMU-ALG' && preceding.querySelector('emu-alg') == null && preceding.textContent != null) {
+      for (let name of preceding.textContent.matchAll(/\b_([a-zA-Z0-9]+)_\b/g)) {
+        scope.declare(name[1], null);
+      }
+    }
+    preceding = previousOrParent(preceding, parentClause);
+  }
+
+  walkAlgorithm(steps, scope, report);
+
+  for (let [name, { kind, used, node }] of scope.vars) {
+    if (!used && node != null && kind !== 'parameter' && kind !== 'abstract closure parameter') {
+      // prettier-ignore
+      let message = `${JSON.stringify(name)} is declared here, but never referred to`;
+      report({
+        ruleId: 'unused-declaration',
+        message,
+        line: node.location.start.line,
+        column: node.location.start.column,
+      });
+    }
+  }
+}
+
+type HasLocation = { location: { start: { line: number; column: number } } };
+type VarKind =
+  | 'parameter'
+  | 'variable'
+  | 'abstract closure parameter'
+  | 'abstract closure capture'
+  | 'loop variable'
+  | 'attribute declaration';
+class Scope {
+  declare vars: Map<string, { kind: VarKind; used: boolean; node: HasLocation | null }>;
+  declare report: Reporter;
+  constructor(report: Reporter) {
+    this.vars = new Map();
+    // TODO remove this when regex state objects become less dumb
+    for (let name of ['captures', 'input', 'startIndex', 'endIndex']) {
+      this.declare(name, null);
+    }
+    this.report = report;
+  }
+
+  declared(name: string): boolean {
+    return this.vars.has(name);
+  }
+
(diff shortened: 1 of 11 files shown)


src/lint/rules/variable-use-def.ts
```diff
@@ -0,0 +1,352 @@
...
+}
+
+type VariableNode = UnderscoreNode & { contents: [TextNode] };
+export function checkVariableUsage(
+  containingAlgorithm: Element,
+  steps: OrderedListNode,
+  report: Reporter
+) {
+  if (containingAlgorithm.hasAttribute('replaces-step')) {
+    // TODO someday lint these by doing the rewrite (conceptually)
```
michaelficarra, line 72: aww, I was looking forward to seeing how you did this

bakkot: It's not _that_ hard - you know exactly the step being replaced, so you can swap out that step in the original and re-run this algorithm, discarding any warnings which don't apply to the replacement section - but it just doesn't come up enough to be worth worrying about.

src/lint/rules/variable-use-def.ts
```diff
@@ -0,0 +1,352 @@
...
+    // handle loops
+    if (first.name === 'text' && first.contents.startsWith('For each ')) {
+      let loopVar = parts[firstRealIndex + 1];
+      if (loopVar?.name === 'pipe') {
+        loopVar = parts[firstRealIndex + 3]; // 2 is the space
+      }
+      if (isVariable(loopVar)) {
+        loopVars.add(loopVar);
+
+        scope.declare(loopVar.contents[0].contents, loopVar, 'loop variable');
```
michaelficarra, line 177: `break` here so you don't try to declare the loop target or anything mentioned in the guard

src/lint/rules/variable-use-def.ts
```diff
@@ -0,0 +1,352 @@
...
+        // check for "Let _x_ be" / "_x_ and _y_ such that"
+        if (i < parts.length - 1) {
+          const next = parts[i + 1];
+          const isSuchThat = next.name === 'text' && next.contents.startsWith(' such that ');
+          const isBe = next.name === 'text' && next.contents.startsWith(' be ');
+
+          if (isSuchThat || isBe) {
+            const varsDeclaredHere = [part];
+            let varIndex = i - 1;
+            // walk backwards collecting this comma/'and' seperated list of variables
```
michaelficarra, line 276: I'd kind of rather get rid of these. We can leave this support for now, but can you open an issue on 262 that lists these cases so we can review them for phrasing?

test/lint-variable-use-def.js
```diff
@@ -0,0 +1,371 @@
...
+        `
+          <emu-alg>
+            1. Let _x_, _y_, and _z_ be some values.
+            1. Return _x_ + _y_ + _z_.
+          </emu-alg>
+        `
+      );
+    });
+
+    it('"such that" counts as a declaration"', async () => {
```
michaelficarra, line 308:
```suggestion
    it('"such that" counts as a use"', async () => {
```

src/lint/rules/variable-use-def.ts
bakkot: Not sure about an issue on 262, but here's a list for you.

- [This one line in JSON.stringify](https://github.com/tc39/ecma262/pull/2900) which should just get fixed
- Module Record [GetBindingValue](https://tc39.es/ecma262/multipage/executable-code-and-execution-contexts.html#sec-module-environment-records-getbindingvalue-n-s) step 3.a
- [Number::toString](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-numeric-types-number-tostring) step 5
- [Number.prototype.toExponential](https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-number.prototype.toexponential) step 10
- [Number.prototype.toPrecision](https://tc39.es/ecma262/#sec-number.prototype.toprecision) step 10

In Temporal there is also [ParseTimeZoneOffsetString](https://tc39.es/proposal-temporal/#sec-temporal-parsetimezoneoffsetstring) step 3.

test/lint-variable-use-def.js
bakkot: Nope, it's a declaration. For example

```
While _finalizationRegistry_.[[Cells]] contains a Record _cell_ such that _cell_.[[WeakRefTarget]] is ~empty~, [...]
```
declares `_cell_`.

src/lint/rules/variable-use-def.ts
bakkot: We don't want to _declare_ anything else, but we do want to count references on the rest of the line as uses, and just putting a `continue` here would skip that. What test are you imagining `continue`ing here would fix?

michaelficarra (approved):

src/lint/rules/variable-use-def.ts
michaelficarra:
```
1. For each integer _a_ of _b_ such that _c_ relates to _a_ in some way, do
```

bakkot: That case is guarded explicitly: https://github.com/tc39/ecmarkup/blob/23e5a62eb15efdb3cb726577ed19d54d19f91c25/src/lint/rules/variable-use-def.ts#L294.

bakkot: There's a test for it too: https://github.com/tc39/ecmarkup/blob/23e5a62eb15efdb3cb726577ed19d54d19f91c25/test/lint-variable-use-def.js#L102-L116

Though not using the "for each" style; I can add one for that case, if you want.

src/lint/rules/variable-use-def.ts
jmdyck: (Note that  [PR #2288](https://github.com/tc39/ecma262/pull/2288) takes care of Module Record GetBindingValue step 3.a.)

src/lint/rules/variable-use-def.ts
michaelficarra: Yes, please.

bakkot: Done.

Issue #488, opened 2022-10-05
HTML character reference processing is inconsistent


gibson042:
#481 introduced processing of named character references, but is inconsistent with the HTML spec (cf. https://html.spec.whatwg.org/multipage/parsing.html#tokenization and specifically https://html.spec.whatwg.org/multipage/parsing.html#named-character-reference-state )—it matches without a terminal semicolon but _also_ fails to handle numeric references or the prefix matching that renders e.g. `I'm &notit; I tell you` as `I'm ¬it; I tell you`, and the inner logic lowercasing `&le;` (which is its own bug and should be `&lt;`) and `&amp;` only with a terminal semicolon is inconsistent with it being optional in the initial match and also inconsistent with the lack of e.g. `AMp;` in https://html.spec.whatwg.org/multipage/named-characters.html#named-character-references .

I think it should instead be either unconditionally strict[^1] or else better aligned with HTML[^2], and in either case should probably also be updated for numeric references and correct case handling.

[^1]: e.g.,
    ```patch
    --- src/formatter/text.ts
    +++ src/formatter/text.ts
    @@ -9,1 +9,1 @@
    -  text = text.replace(/&[a-zA-Z0-9]+;?/g, m => {
    +  text = text.replace(/&[a-zA-Z0-9]+;/g, m => {
    ```
[^2]: e.g.,
    ```patch
    --- src/formatter/text.ts
    +++ src/formatter/text.ts
    @@ -11,2 +11,4 @@
    -    if ({}.hasOwnProperty.call(entities, m) && entities[m] !== null) {
    -      return entities[m];
    +    const prefix = Array(m.length).fill().map((_, n) => n ? m.slice(0, -n) : m)
    +      .find(prefix => ({}).hasOwnProperty.call(entities, prefix) && entities[prefix] !== null);
    +    if (prefix) {
    +      return entities[prefix] + m.slice(prefix.length);
    ```

bakkot:
> matches without a terminal semicolon

Only for legacy entities, which is consistent with HTML. The list of entities matched is taken directly from [the HTML spec](https://github.com/whatwg/html-build/tree/main/entities).

> _also_ fails to handle numeric references

I don't think we have any uses of that form to check if the decision to use that form was deliberate (e.g. the numeric value of the code point is relevant to the reader), so I didn't implement it. If there's some usages out there which you think are definitely not deliberate, I'd be happy to accept a PR adding that.

[...]

PR #490, opened 2022-10-06
formatter: render numeric character references like named ones

gibson042:
Ref #488

### DIFF
--- a/src/formatter/text.ts
+++ b/src/formatter/text.ts
@@ -1,12 +1,50 @@
 import { LineBuilder } from './line-builder';
 const entities = require('../../entities-processed.json');
 
+function isBadNumericReference(codePoint: number) {
+  // https://html.spec.whatwg.org/multipage/parsing.html#numeric-character-reference-end-state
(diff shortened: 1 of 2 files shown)


bakkot:
This will render some things it probably shouldn't - e.g. `&#xA0;` (no-break space), which is visually indistinguishable from a regular space character in most editors.

I'd like to filter on [the same regex as we use for regular entities](https://github.com/tc39/ecmarkup/blob/7bc7f2eff2943ae6f3d7918f9e3b4a004c88f950/filter-entities.js#L14): i.e. don't render entities which match `/\p{White_Space}|\p{DI}|\p{gc=M}|\p{gc=C}/u`. I can push a commit with that change sometime if you don't get to it before me.

(It might actually be nice to translate numeric forms of entities which we want to keep in entity form into their corresponding named entity, when they have a name - so `&#A0;` into `&nbsp;` - but that doesn't need to happen in this PR.)

bakkot (approved):

PR #500, opened 2023-01-10
Clean up documentation

gibson042:
* Fix spelling of "separated"/"separating"
* Remove duplicate lines
* Use correct case for documenting attributes
* Fix gaps and errors in attribute documentation
* Remove erroneous mention of content constraints in emu-nt
* Clean up descriptions
* Add more structure to emu-alg documentation
* Clean up effects propagation documentation
* Enable effects propagation in documentation
* Clean up documentation examples

The most convenient way to review is probably commit-by-commit with whitespace changes suppressed.

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -45,10 +45,10 @@ ecmarkup spec.html out.html
       <tr><th>Option</th><th>Description</th></tr>
       <tr><td>watch</td><td>Rebuild when files change</td></tr>
       <tr><td>verbose</td><td>Print verbose logging info</td></tr>
-      <tr><td>write-biblio</td><td>Emit a biblio file to the specified location</td></tr>
-      <tr><td>assets</td><td>"none" for no css/js, "inline" for inline css/js</td></tr>
-      <tr><td>css-out</td><td>Emit the Ecmarkup CSS file to the specified location</td></tr>
-      <tr><td>js-out</td><td>Emit the Ecmarkup JS file to the specified location</td></tr>
+      <tr><td>write-biblio</td><td>Emit a biblio file to the specified path</td></tr>
+      <tr><td>assets</td><td>"none" for no CSS/JS, "inline" for inline CSS/JS</td></tr>
+      <tr><td>css-out</td><td>Emit the Ecmarkup CSS file to the specified path</td></tr>
+      <tr><td>js-out</td><td>Emit the Ecmarkup JS file to the specified path</td></tr>
       <tr><td>lint-spec</td><td>Enforce some style and correctness checks</td></tr>
       <tr><td>lint-formatter</td><td>The <a href="https://eslint.org/docs/user-guide/formatters/">eslint formatter</a> to be used for printing warnings and errors when using --verbose. Either the name of a built-in eslint formatter or the package name of an installed eslint compatible formatter.</td></tr>
       <tr><td>strict</td><td>Exit with an error if there are warnings. Cannot be used with --watch</td></tr>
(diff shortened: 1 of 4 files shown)


spec/index.html
```diff
@@ -31,54 +32,57 @@ <h1>Ecmarkup</h1>
...
+      <tr><td>`--assets`</td><td>`assets`</td><td>"none" for no CSS/JS, "inline" for inline CSS/JS.</td></tr>
+      <tr><td>`--css-out`</td><td>`cssOut`</td><td>Emit the Ecmarkup CSS file to the specified path.</td></tr>
+      <tr><td>`--js-out`</td><td>`jsOut`</td><td>Emit the Ecmarkup JS file to the specified path.</td></tr>
+      <tr><td>`--lint-spec`</td><td>`lintSpec`</td><td>Enforce some style and correctness checks.</td></tr>
+      <tr><td>`--error-formatter`</td><td></td><td>The <a href="https://eslint.org/docs/user-guide/formatters/">eslint formatter</a> to be used for printing warnings and errors when using `--verbose`. Either the name of a built-in eslint formatter or the package name of an installed eslint compatible formatter.</td></tr>
+      <tr><td>`--strict`</td><td></td><td>Exit with an error if there are warnings. Cannot be used with `--watch`.</td></tr>
+      <tr><td>`--mark-effects`</td><td>`markEffects`</td><td>Propagate and render <a href="#effects">effects</a> like "user code".</td></tr>
+      <tr><td>`--multipage`</td><td></td><td>Emit a distinct page for each top-level clause.</td></tr>
```
gibson042, line 58: I'm not clear on the difference between "Build options" vs. "Document options", but it does seem strange for `markEffects` to not be in the same category as `toc`.

spec/index.html
```diff
@@ -503,282 +556,330 @@ <h2>Example</h2>
...
-    <p><b>Params:</b> Parameters for this production. Multiple parameters separated by commas.</p>
-    <p><b>Type:</b> Type of production, either "lexical" or "regexp". Default is blank (normal).</p>
-    <p><b>oneof:</b> If present, production is a one-of production. See DecimalDigit example above.</p>
-    <p><b>collapsed:</b> If present, production is displayed in collapsed format with LHS and RHS on the same line.</p>
+    <p><b>name:</b> Required: Name of the production (i.e. the LHS). Should conform with the syntax of <a href="https://www.unicode.org/reports/tr31/#D1">Unicode Standard Annex #31, UAX31-D1 Default Identifier Syntax</a> and not include any underscore (U+005F LOW LINE “<b>_</b>”).</p>
+    <p><b>collapsed:</b> If present, the production is displayed in collapsed format with LHS and RHS on the same line.</p>
+    <p><b>oneof:</b> If present, the production is a one-of production. See the DecimalDigit example <emu-xref href="#grammar">above</emu-xref>.</p>
+    <!-- what does it even mean for a LHS to be optional??? -->
+    <p><b>optional:</b> If present, the LHS is marked as optional.</p>
```
gibson042, line 775: I think this attribute should just be removed, but maybe there's a case I'm just not getting for
```html
<emu-production>
  <emu-nt optional="">LHS</emu-nt> <emu-geq>:</emu-geq>
  <emu-rhs>…</emu-rhs>
  …
</emu-production>
```

spec/index.html
```diff
@@ -503,282 +556,330 @@ <h2>Example</h2>
...
+    <p><b>params:</b> Parameters for this production. Multiple parameters are separated by commas with optional whitespace.</p>
+    <p><b>primary:</b> Optional: The production is authoritative and should be the target of references.</p>
+    <p><b>type:</b> Optional: Type of production. If absent, the production is considered syntactic.</p>
+    <ul>
+      <li>"lexical": The production is part of a lexical grammar, and should render with the LHS and RHS separated by two colons “<b>::</b>”.</li>
+      <li>"regexp": Deprecated because the ECMAScript regular expression grammar is considered to be lexical. Productions with this type render with the LHS and RHS separated by three colons “<b>:::</b>”.</li>
+    </ul>
```
gibson042, line 782: I don't like having default values that are not theirselves part of the enumeration... should we just add support for e.g. `type="syntactic"`?

spec/index.html
bakkot: I'm not entirely clear on the difference either, but I agree `markEffects` probably ought to be under "document options".

spec/index.html
bakkot: Yeah I have no idea what that's for.

Also, I'm not sure it makes sense to support writing `emu-production` (etc) manually; the happy path is using grammarkdown notation. Possibly these should not be in the documentation at all. (They'll still be implicitly supported because grammarkdown emits these nodes, but we don't need to tell people about them, I think.)

spec/index.html
bakkot: I understand that preference, but I prefer to have only one way of writing anything, when possible, so I would prefer to avoid adding an alias here.

bakkot (approved):
This is great, thank you!

PR #504, opened 2023-01-11
Match synthesized selectable counter markers to the context list

### LINKED ISSUE #503: Synthesized counter markers are incorrect for replaced steps
[listNumbers.js](https://github.com/tc39/ecmarkup/blob/a004ca7d8e95452e680e284c9a4a86a0d31e9914/js/listNumbers.js) ignores `start` and nesting overrides, as observable in the HTML that replaces https://tc39.es/ecmarkup/#step-replace-me:
```html
<emu-alg replaces-step="step-replace-me">
  <ol start="1" class="nested-once">
    <li><span aria-hidden="true" style="font-size: 0px;">1. </span>Replacement.<ol>
      <li><span aria-hidden="true" style="font-size: 0px;">a. </span>Substep.</li>
    </ol></li>
  </ol>
</emu-alg>
```

Emphasizing the incorrect generated spans (which should be "**a.**" and "**i.**"):

<blockquote>
  <ol start="1" class="nested-once" type="lower-alpha">
    <li><strong>1. </strong>Replacement.<ol type="lower-roman">
      <li><strong>a. </strong>Substep.</li>
    </ol></li>
  </ol>
</blockquote>

gibson042:
Fixes #503


bakkot (approved):
Confirmed this works locally, and thanks for doing the work of implementing these correctly instead of just hardcoding a few cases like I did.

PR #505, opened 2023-01-18
enforce own spec is formatted in CI

bakkot:


package.json
```diff
@@ -9,6 +9,7 @@
 		"build-release": "tsc",
 		"build-spec": "mkdir -p docs && node bin/ecmarkup.js spec/index.html docs/index.html --css-out docs/elements.css --js-out docs/ecmarkup.js",
 		"prepack": "safe-publish-latest && npm run build-release",
+    "format": "emu-format --write spec/index.html",
```
ljharb, line 12:
```suggestion
		"format": "emu-format --write spec/index.html",
```

bakkot: Oh no tabs :(

PR #506, opened 2023-01-18
Lowercase getAttribute arguments to match emu-format

gibson042:
Ref 946c5c156f9488ffc55a7b8bd741437d39d7052b


bakkot (approved):

bakkot:
Not just emu-format, btw; [the spec says](https://html.spec.whatwg.org/multipage/dom.html#attr-data-*) they're not supposed to contain any ASCII upper chars.

PR #512, opened 2023-01-21
Initialize to a reasonable amount of TOC expansion

gibson042:
Many proposals have a small TOC, especially regarding the number of top-level items, for example:
* https://tc39.es/proposal-atomics-wait-async/
* https://tc39.es/proposal-shadowrealm/
* https://tc39.es/proposal-change-array-by-copy/
* https://tc39.es/proposal-json-parse-with-source/

This PR improves the initialization of such TOCs to include automatic expansion up to an arbitrary limit of 20 total visible items, without affecting large TOCs and/or reloading of saved state.

Auto-expansion is limited to full breadth at a given depth (to avoid asymmetric expansion of low-numbered items), and there is an exception allowing expansion of single-item depth levels to exceed the limit (for cases similar to https://tc39.es/proposal-atomics-wait-async/ , in which expansion of the only substantive top-level item should not be suppressed even if there are many items underneath it).


gibson042:
I'm fine with keeping the animation.

bakkot (approved):

bakkot (approved):

bakkot:
Sorry the delay. I do find the animation a bit distracting, but that can get fixed later.

PR #517, opened 2023-02-01
Dynamically add content to make <sup>s copy-paste safe

### LINKED ISSUE #516: Superscript text representing exponents is not copy-paste safe
Elements such as `<sup>` in data copied from spec content are replaced with their textContent when pasting as plain text, which can result in semantically inappropriate transformations such as (plucking real-word cases from [Number.prototype.toFixed](https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-number.prototype.tofixed)) "If x ≥ 10<sup>21</sup>" → "If x ≥ 1021" and "n / 10<sup>f</sup> - x" → "n / 10f - x".

This is a common need when writing test262 tests (cf. https://github.com/tc39/test262/blob/main/CONTRIBUTING.md#info ), so we should try to improve the experience similarly to what was done for step labels in #290.

gibson042:
Fixes #516

* Replace a simple numeric text `<sup>` with Unicode superscript characters.
* Wrap a `<sup>` having other text and/or `<var>` contents with hidden copyable text (a leading [`↑`](https://en.wikipedia.org/wiki/Knuth%27s_up-arrow_notation) and [iff necessary] parentheses).
* Leave other `<sup>`s alone (to not break e.g. `1<sup>st</sup>`).

### DIFF
--- a/js/superscripts.js
+++ b/js/superscripts.js
@@ -1,9 +1,16 @@
 'use strict';
 
 // Update superscripts to not suffer misinterpretation when copied and pasted as plain text.
-// For example, `10<sup>3</sup>` becomes `103`, but `10³` is preserved.
+// For example,
+// * Replace `10<sup>3</sup>` with `10³` so it gets pasted as that rather than `103`.
+// * Replace `10<sup>-<var>x</var></sup>` with
+//   `10<span aria-hidden="true"> ↑ </span><sup>-<var>x</var></sup>`
+//   so it gets pasted as `10 ↑ -x` rather than `10-x`.
+// * Replace `2<sup><var>a</var> + 1</sup>` with
+//   `2<span …> ↑ (</span><sup><var>a</var> + 1</sup><span …>)</span>`
+//   so it gets pasted as `2 ↑ (a + 1)` rather than `2a + 1`.
 
-const arithmeticReplacements = new Map([
+const supReplacements = new Map([
   ['0', '⁰'],
   ['1', '¹'],
   ['2', '²'],
@@ -23,25 +30,27 @@ const arithmeticReplacements = new Map([
   [' ', ' '],
 ]);
 
-function makeSuperscriptAccessible(sup) {
-  // Replace a text-only superscript with text when it is fully replaceable,
+function makeExponentPlainTextSafe(sup) {
+  // Replace a text-only <sup> if its contents are fully representable,
   // but otherwise do not modify it (for cases such as `1<sup>st</sup>`).
   if ([...sup.childNodes].every(node => node.nodeType === 3)) {
-    const replacementText = sup.textContent
+    const text = sup.textContent;
+    const replacementText = text
       .split('')
-      .map(ch => arithmeticReplacements.get(ch) || '')
+      .map(ch => supReplacements.get(ch) || '')
       .join('');
-    if (replacementText.length === sup.textContent.length) {
+    if (replacementText.length === text.length) {
       sup.replaceWith(replacementText);
     }
     return;
   }
-  // Wrap a not-yet-handled superscript with hidden Knuth up-arrow notation
-  // if it appears to represent exponentiation.
-  if (!sup.classList.contains('text') && !sup.querySelector('*:not(var)')) {
+
+  // Add a hidden Knuth up-arrow before a <sup> that appears to represent an exponent.
+  if (!sup.querySelector('*:not(var)')) {
     let prefix = ' ↑ ';
     let suffix = '';
-    if (sup.childNodes.length > 1 && !/^\(.*\)$/s.test(sup.textContent.trim())) {
+    // Add wrapping parentheses unless they are already present or there are no medial spaces.
+    if (!/^\(.*\)$|^\S*$/s.test(sup.textContent.trim())) {
       prefix += '(';
       suffix += ')';
     }
@@ -53,7 +62,7 @@ function makeSuperscriptAccessible(sup) {
 }
 
 document.addEventListener('DOMContentLoaded', () => {
-  document.querySelectorAll('sup').forEach(sup => {
-    makeSuperscriptAccessible(sup);
+  document.querySelectorAll('sup:not(.text)').forEach(sup => {
+    makeExponentPlainTextSafe(sup);
   });
 });
--- a/js/superscripts.js
+++ b/js/superscripts.js
@@ -1,0 +1,59 @@
+'use strict';
+
+// Update superscripts to not suffer misinterpretation when copied and pasted as plain text.
+// For example, `10<sup>3</sup>` becomes `103`, but `10³` is preserved.
+
+const arithmeticReplacements = new Map([
+  ['0', '⁰'],
+  ['1', '¹'],
+  ['2', '²'],
+  ['3', '³'],
+  ['4', '⁴'],
+  ['5', '⁵'],
+  ['6', '⁶'],
+  ['7', '⁷'],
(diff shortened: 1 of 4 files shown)


js/superscripts.js
```diff
@@ -0,0 +1,68 @@
...
+  ['5', '⁵'],
+  ['6', '⁶'],
+  ['7', '⁷'],
+  ['8', '⁸'],
+  ['9', '⁹'],
+  ['(', '⁽'],
+  [')', '⁾'],
+  ['=', '⁼'],
+  ['+', '⁺'],
+  ['-', '⁻'],
```
bakkot, line 28: There are visual equivalents for multiplication and division as well:
```suggestion
  ['-', '⁻'],
  ['×', 'ˣ'],
  ['/', 'ᐟ'],
```

though I suppose it's considered poor form to abuse Unicode in this way (https://github.com/tc39/ecmarkup/pull/265#issuecomment-722978302).

(Those are U+02e3 "Modifier Letter Small X" and U+141F "Canadian Syllabics Final Acute" respectively.)

Why does Unicode have superscript addition but not multiplication, anyway?

gibson042: :shrug:

gibson042: > There are visual equivalents for multiplication and division as well:
> 
> though I suppose it's considered poor form to abuse Unicode in this way ([#265 (comment)](https://github.com/tc39/ecmarkup/pull/265#issuecomment-722978302)).
> 
> (Those are U+02e3 "Modifier Letter Small X" and U+141F "Canadian Syllabics Final Acute" respectively.)

Yeah, I definitely don't want to misuse characters like that—preserving `10<sup>1 × 3</sup>` is better than replacing it with `10¹ ˣ ³`. If the inconsistency between `+` and `×` bothers you, I'd rather just remove the former (but note that use as addition in variable-free text is unexpected; it's included here to cover sign assertions such as `10<sup>+3</sup>` → `10⁺³`, which don't currently exist in ECMA-262 or ECMA-402).

> Why does Unicode have superscript addition but not multiplication, anyway?

:shrug:

michaelficarra:
I don't want to solve this problem this way. It seems like a problem that should be solved by the UA (if at all), not each individual web page that uses `<sup>`.

gibson042:
Maybe so, but `<sup>` is used for more than just exponentiation. And doing nothing does have consequences that require people to know about this issue: https://github.com/search?q=repo%3Atc39%2Ftest262+%2F%5B*%C3%97%5D+10%28%5B2-9%5D%7C%5B13-9%5D%5B0-9%5D%29%5Cb%2F&type=code

michaelficarra:
Yeah that still doesn't make it our problem to solve.

gibson042:
How is it categorically different from the "_\<li> numbers are not included in copied text_" problem, which ecmarkup **does** address?

michaelficarra:
There's an easy, straightforward, generic solution to the step numbering issue (though it does have its issues). Maybe I would be more open to solving the exponentiation issue with something like a leading `^(` and a trailing `)`.

gibson042:
> Maybe I would be more open to solving the exponentiation issue with something like a leading `^(` and a trailing `)`.

Like this PR currently does (_no `↑`/`^` before detected non-exponentiation and no parentheses around a simple exponent_), or in all cases?
1. source `1<sup>st</sup>`; "1st" vs. "1^(st)"
2. source `10<sup>3</sup>`; "10³" vs. "10^3" vs. "10^(3)"
3. source `10<sup>-<var>x</var></sup>`; "10^(-x)"
4. source `2<sup><var>a</var> + 1</sup>`; "2^(a + 1)"

bakkot:
Per editor call:

- while the unicode characters are neat, they're more complexity than warranted; just adding the exponentiation symbol and (if necessary) parens in all (relevant) cases works fine
  - i.e., `x<sup>2</sup>` becomes `x**2`, `x<sup>i + 1</sup>` becomes `x**(i + 1)`
- Knuth's up arrow is too obscure; either `^` (the usual convention, but conflicts with JS xor) or `**` (what JS uses) would work

We didn't talk about the precise heuristic for "things that look like exponentiation" but using some heuristic here (as you currently do) is fine; I'll review the precise heuristic you have later but I expect it's fine.

gibson042:
> * i.e., `x<sup>2</sup>` becomes `x**2`, `x<sup>i + 1</sup>` becomes `x**(i + 1)`

@bakkot Done.

js/superscripts.js
```diff
@@ -0,0 +1,54 @@
...
+
+function makeExponentPlainTextSafe(sup) {
+  // Change a <sup> only if it appears to be an exponent:
+  // * text-only and contains only mathematical content (not e.g. `1<sup>st</sup>`)
+  // * contains only <var>s and internal links (e.g.
+  //   `2<sup><emu-xref><a href="#ℝ">ℝ</a></emu-xref>(_y_)</sup>`)
+  const isText = [...sup.childNodes].every(node => node.nodeType === 3);
+  const text = sup.textContent;
+  if (isText) {
+    if (!/^[0-9. 𝔽ℝℤ()=*×/÷±+−-]+$/u.test(text)) {
```
michaelficarra, line 23:
```suggestion
    if (!/^[0-9. 𝔽ℝℤ()=*×/÷±+\u2212-]+$/u.test(text)) {
```

I wouldn't include anything that's a confusable for a RegExp control character in that particular context. This would include `-`, `\`, `^`, and `]` in character classes, for example.

gibson042: Fixed.

michaelficarra: @gibson042 There's another one below.

gibson042: Also fixed. :facepalm:

js/superscripts.js
```diff
@@ -0,0 +1,55 @@
...
+    }
+  }
+
+  let prefix = '**';
+  let suffix = '';
+
+  // Add wrapping parentheses unless they are already present
+  // or this is a simple (possibly signed) integer or single-variable exponent.
+  const skipParens =
+    /^\(.*\)$/s.test(text.trim()) ||
```
bakkot, line 38: This will incorrectly omit parens for e.g. `<sup>(a+1)*(b+1)</sup>` but probably not worth worrying about.

bakkot (approved):
LGTM, thanks!

js/superscripts.js
gibson042: Agreed.

gibson042: Although we _could_ fix it without too much trouble, just at the cost of simplicity:
```suggestion
    /^[±+\u2212-]?(?:[0-9]+|\p{ID_Start}\p{ID_Continue}*)$/u.test(text) ||
    // Split on parentheses and remember them; the resulting parts must
    // start and end empty (i.e., with open/close parentheses)
    // and increase depth to 1 only at the first parenthesis
    // to e.g. wrap `(a+1)*(b+1)` but not `((a+1)*(b+1))`.
    text.trim().split(/([()])/g).reduce((depth, s, i, parts) => {
      if (s === "(") {
        return depth > 0 || i === 1 ? depth + 1 : NaN;
      } else if (s === ")") {
        return depth > 0 ? depth - 1 : NaN;
      } else if (s === "" || i > 0 && i < parts.length - 1) {
        return depth;
      }
      return NaN;
    }, 0) === 0;
```

bakkot: Since you've already written it, might as well include it. Pushed it up and will land with this added.

PR #522, opened 2023-04-04
Feat(style): match with tc 39 website

AugustinMauroy:


css/elements.css
```diff
@@ -1,9 +1,11 @@
+@import url(https://fonts.googleapis.com/css2?family=Open+Sans:wght@300;400;600&amp;family=Source+Sans+Pro:wght@300;400;600&amp;display=swap);
```
ljharb, line 1: i'm not sure we want a live dependency on google's font API

AugustinMauroy: Do you think that having the font downloaded in the documents would be better?

bakkot:
Thanks for the suggestion, but I don't think we want to change the styling at this time.

css/elements.css
ljharb: Certainly hosting it ourselves would be better, but i'm not sure how that'd work for all the proposal repos this serves. What's wrong with the current font for specs?

PR #524, opened 2023-04-20
Lint for consistency in multiline vs single-line "if" and "else"

bakkot:
See https://github.com/tc39/ecma262/pull/3043.

Needed some minor refactoring to have enough context to apply the rule, so I recommend reviewing commits individually.

Calling this major due to adding a new on-by-default lint rule, which will break the build of most nontrivial specs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "ecmarkup",
-	"version": "16.2.0",
+	"version": "17.0.0",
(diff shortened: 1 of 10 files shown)


michaelficarra (approved):
Code LGTM but could use a few more tests. Two negative tests for if/else without else-if. And as @jmdyck points out, positive tests where if and else-if do not agree.

PR #526, opened 2023-04-29
better heuristic for determining active clause

michaelficarra:
I got frustrated by the TOC not always correctly determining the clause that I was reading. Experimentally, this heuristic works very well. It chooses either the first clause that is fully visible (except when at the bottom of the viewport) or, if there are none, the clause that straddles the vertical midpoint of the viewport. This has the helpful property that clicking the TOC item always ends up marking that item as active (except possibly the very final clause(s) of the document).

*edit:* For reference, the previous heuristic was that the clause straddled the top of the viewport, meaning the active clause was often mostly out of view.


js/menu.js
```diff
@@ -367,25 +367,64 @@ Menu.prototype.revealInToc = function (path) {
...
+      (clauseTop > 0 && clauseTop < window.innerHeight) ||
+      (clauseBottom > 0 && clauseBottom < window.innerHeight) ||
+      (clauseTop < 0 && clauseBottom > window.innerHeight);
+
+    if (isPartiallyVisible) {
+      let innerPath = path.concat($clause);
+      result.push([$clause, innerPath]);
+      result.push(...getVisibleClauses($clause, innerPath));
+    }
+  }
```
michaelficarra, line 423: I guess we could early exit the loop here after first seeing visible clauses and then seeing no-longer visible clauses.

michaelficarra: I did this.

bakkot (approved):

PR #533, opened 2023-06-18
Fix <ins>/<del> with structured headers

### LINKED ISSUE #530: <ins>/<del> don't work with structured headers
There's an example in [test/baselines/sources/structured-headers.html](https://github.com/tc39/ecmarkup/blob/879aa2474d221c126748c84d79004a7f621b5172/test/baselines/sources/structured-headers.html#L39-L50), but it generates a warning and so doesn't work with `--strict`.

### LINKED ISSUE #531: "clause numbers should be strictly increasing" logic is incorrect
[test/baselines/sources/clauses.html](https://github.com/tc39/ecmarkup/blob/879aa2474d221c126748c84d79004a7f621b5172/test/baselines/sources/clauses.html#L51) generates a warning.

gibson042:
Fixes #530
Fixes #531

Supports \<ins>/\<del> both inside and outside the structured-header \<h1>.
Also extends baseline testing to cover warnings, and adds hopefully-temporary suppression of warnings that seem unexpected to me.


test/baselines/sources/structured-headers.html
```diff
@@ -23,6 +23,8 @@ <h1>
   </emu-alg>
 </emu-clause>
 
+<!-- FIXME: Shouldn't this warning be suppressed by `redefinition`? -->
```
bakkot, line 26: No - `id`s have effects on the document considered as a piece of HTML (i.e., for computers), not just on the document considered as a piece of writing (i.e., for humans).

"redefinition" means that we are redefining an algorithm, which is normally an error when considering the document as a piece of writing (since a human reader expects a given term to have only one definition). But it doesn't have any interaction with the error of reusing an `id`, since that's a problem for the _computer_, not just for humans: the computer needs distinct `id`s for each clause so that you can link to them independently.

test/baselines/generated-reference/namespaces.html
```diff
@@ -44,6 +44,8 @@ <h1><span class="secnum">1.1</span> Clause 1.1</h1>
 </emu-production>
 </emu-grammar>
 
+    <!-- FIXME: Shouldn't this be allowed by the preceding `namespace=clause`? -->
```
bakkot, line 47: I wouldn't think so - line 53 has an algorithm with an identical AOID, and is within the same namespace.

test/baselines/generated-reference/namespaces.html
```diff
@@ -63,10 +65,14 @@ <h1><span class="secnum">A</span> Annex</h1>
     </emu-rhs>
 </emu-production>
 </emu-grammar>
+  <!-- FIXME: Shouldn't this be allowed by the preceding `namespace=annex`? -->
+  <!-- EXPECT_WARNING { "ruleId": "duplicate-definition", "message": "duplicate definition \"SomeAlg\"" } -->
   <emu-annex id="annex11" aoid="SomeAlg">
     <h1><span class="secnum">A.1</span> SomeAlg</h1>
   </emu-annex>
 
+  <!-- FIXME: Shouldn't this be allowed by `namespace=annex2`? -->
```
bakkot, line 74: I'm not sure, but I would assume that namespaces apply to the contents of a clause/algorithm, not to the AOID of the algorithm itself.

src/utils.ts
```diff
@@ -117,6 +117,33 @@ export function replaceTextNode(node: Node, frag: DocumentFragment) {
...
+  | 'firstChild'
+  | 'firstElementChild'
+  | 'lastChild'
+  | 'lastElementChild';
+
+/*@internal*/
+export function traverseWhile(
+  node: Node | null,
+  relationship: NodeRelationship,
+  cb: (node: Node) => boolean,
```
bakkot, line 136:
```suggestion
  predicate: (node: Node) => boolean,
```

bakkot:
Thanks!

I pushed [a commit](https://github.com/tc39/ecmarkup/pull/533/commits/0ea5b748b5c3771583b6462a21b01e895ba823ac) making the types for `traverseWhile` more precise, to avoid casts, and a [second](https://github.com/tc39/ecmarkup/pull/533/commits/65da20bcc2cc6dd8ec8179e183023eff24e01ac2) addressing my other comments. LGTM now.

bakkot:
Published in v17.0.1.

justingrant:
I just tried to use this and got an error.  Expected?

Spec text: 

```
      <h1>
        <del>IsTimeZoneOffsetString</del><ins>IsOffsetTimeZoneIdentifier</ins> (
          _offsetString_: a String,
        ): a Boolean
      </h1>
```

Result: 
`Error: spec/mainadditions.html:444:21: could not find matching del tag`

gibson042:
Hmm, I don't know if ecmarkup should support that level of granularity. But if it does, then the starting point would look something like this (which notably swallows the old name, a characteristic that is probably not desirable):
```diff
--- src/header-parser.ts
+++ src/header-parser.ts
@@ -70,7 +70,16 @@ function parseH1(headerText) {
     return { type: 'failure', errors };
   }
   offset += match[0].length;
-  const name = match[0].trimRight();
+  let name = match[0].trimRight();
+  if (wrappingTag !== null && name.includes(`</${wrappingTag}>`)) {
+    // correctly handle text like `<del>OldName</del><ins>NewName</ins>` as non-wrapped
+    wrappingTag = null;
+    const nameSource = `<${wrappingTag}>${name}`;
+    match = nameSource.match(/<ins[^>]*>([^<]+?)<\/ins>/) || nameSource.match(/>([^<]+?)</);
+    if (match) {
+      name = match[1];
+    }
+  }
 
   if (text === '') {
     if (wrappingTag !== null) {
```

PR #539, opened 2023-07-06
improve accessibility of text

michaelficarra:
<details>
<summary>before:</summary>

</details>

<details>
<summary>after:</summary>

</details>

TODO:
- [x] we should probably do the same for the other fonts
- [ ] ~turn off some optional ligatures?~
- [x] use SVGs or something for the little TOC arrows
- [ ] update the 262 colophon
- [ ] update the description of the styling of enums in `#sec-enum-specification-type` in 262


bakkot (changes requested):
This:

- causes text to move when clicking to highlight a variable
- draws a huge amount of attention to enum values, via a box shadow and a different background color
- makes the highlight colors worse and unnecessarily different from Bikeshed

I don't like any of those and won't merge it while it has those features.

bakkot:
we should lint to make constants kebab-case

bakkot:
Rebased, should be good to go now.

PR #543, opened 2023-08-12
remove --js-out/--css-out; replace with --assets-dir

bakkot:
This is a breaking change. It significantly simplifies things as we introduce more assets, e.g. in https://github.com/tc39/ecmarkup/pull/539.

Had to update typescript to fix some issues with the older version, which required some other related changes. I suggest reviewing commits individually.

### DIFF
--- a/src/Spec.ts
+++ b/src/Spec.ts
@@ -18,7 +18,8 @@ import * as yaml from 'js-yaml';
 import * as utils from './utils';
 import * as hljs from 'highlight.js';
 // Builders
-import Import, { EmuImportElement } from './Import';
+import type { EmuImportElement } from './Import';
+import Import from './Import';
 import Clause from './Clause';
 import ClauseNumbers from './clauseNums';
 import Algorithm from './Algorithm';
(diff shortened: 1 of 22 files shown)


src/Spec.ts
```diff
@@ -562,12 +584,12 @@ export default class Spec {
       (await concatJs(sdoJs, tocJs)) + `\n;let usesMultipage = ${!!this.opts.multipage}`;
     const jsSha = sha(jsContents);
 
+    await this.buildAssets(jsContents, jsSha);
+
     if (this.opts.multipage) {
-      await this.buildMultipage(wrapper, commonEles, jsSha);
+      await this.buildMultipage(wrapper, commonEles);
```
bakkot, line 590: To explain the change to do `buildAssets` before `buildMultipage`:

`buildMultipage` makes a copy of the document as it exists when `buildMultipage` is invoked. Previously we were inserting `ecmarkup.js` etc separately for the two kinds of output: we'd add `script src='../ecmarkup.js'` in `buildMultipage`, and then later add `script src='ecmarkup.js'` in `buildAssets`.

Now we put all assets in the single-page version initially and then rewrite those references in the copy in `buildMultipage` like we rewrite any other reference.

src/Spec.ts
```diff
@@ -1202,63 +1229,50 @@ ${await utils.readFile(path.join(__dirname, '../js/multipage.js'))}
...
+    if (this.assets.type === 'external') {
+      const outDir = this.opts.outfile
+        ? this.opts.multipage
+          ? this.opts.outfile
+          : path.dirname(this.opts.outfile)
+        : process.cwd();
 
-    if (this.opts.assets === 'none') return;
+      const scriptLocationOnDisk = path.join(this.assets.directory, 'ecmarkup.js');
+      const styleLocationonDisk = path.join(this.assets.directory, 'ecmarkup.css');
```
michaelficarra, line 1258:
```suggestion
      const styleLocationOnDisk = path.join(this.assets.directory, 'ecmarkup.css');
```

and elsewhere

michaelficarra (approved):

PR #545, opened 2023-08-28
force page background colour to white

michaelficarra:
Fixes #244. If a user has their default UA background colour set to something dark, the body text may be unreadable. This sidesteps the issue by just setting the background colour to white.


bakkot (approved):
Sure.

PR #546, opened 2023-08-29
allow references pane to be resized

michaelficarra:


js/menu.js
```diff
@@ -840,6 +844,29 @@ let referencePane = {
...
+    let startingMousePos = pointerDownEvent.clientY;
+    let startingHeight = parseInt(getComputedStyle(this.$container).height);
+    let moveListener = pointerMoveEvent => {
+      if (pointerMoveEvent.buttons === 0) {
+        removeListeners();
+        return;
+      }
+      let desiredHeight = startingHeight - (pointerMoveEvent.clientY - startingMousePos);
+      this.$container.style.height =
+        Math.max(50, Math.min(visualViewport.height - 150, desiredHeight)) + 'px';
```
gibson042, line 858: This clamping should probably just be handled by CSS `min-height` and `max-height`.

gibson042 (approved):
I like this! Can you do the same for the menu pane as well?

js/menu.js
michaelficarra: Done.

michaelficarra:
@gibson042 Maybe some other time. Resizing the TOC doesn't seem all that useful to me.

gibson042:
> @gibson042 Maybe some other time. Resizing the TOC doesn't seem all that useful to me.

https://tc39.es/proposal-temporal/#sec-properties-of-the-temporal-plaindatetime-prototype-object
screenshot of Temporal menu pane showing ellipses hiding Temporal.PlainDateTime.prototype property names

michaelficarra:
@gibson042 Yeah but there might be other solutions, like allowing horizontal scroll in the TOC

image

gibson042:
I'd want the pane to be resizable with or without horizontal scrolling.

michaelficarra:
@gibson042 That's fair. You should open an issue to track it.

gibson042:
Or you can add it to this PR, which doesn't have an associated issue AFAICT.

michaelficarra:
@gibson042 I would prefer to have this PR only address the references pane. I ask that you open an issue to track the TOC usability so that, when this PR is merged, we do not forget about the issue you raised.

gibson042:
#549

bakkot (approved):

PR #548, opened 2023-08-30
lint against non-kebab-cased enum values

### LINKED ISSUE #544: consider linting against camelCase enum values
`~enums~` are written with inconsistent casing within and across specs. https://github.com/tc39/ecmarkup/pull/539 will change the styling to smallcaps, which can look a little bit weird with capital letters, so that might be a good impetus to start enforcing `~kebab-case~` enum values everywhere.

cc @tc39/ecma402-editors because this would entail a change in 402 (which has some kebab-case, some camelCase, and some enum values with spaces).

bakkot:
Fixes #544.

### DIFF
--- a/src/lint/rules/enum-casing.ts
+++ b/src/lint/rules/enum-casing.ts
@@ -1,0 +1,25 @@
+import type { Reporter } from '../algorithm-error-reporter-type';
+import type { OrderedListItemNode } from 'ecmarkdown';
+import { offsetToLineAndColumn } from '../../utils';
+
+const ruleId = 'enum-casing';
+
+/*
+Checks that ~enum-values~ are kebab-cased.
+*/
+export default function (report: Reporter, step: OrderedListItemNode, algorithmSource: string) {
+  for (const item of step.contents) {
+    if (item.name !== 'tilde' || item.contents.length !== 1 || item.contents[0].name !== 'text') {
+      continue;
+    }
+    const text = item.contents[0];
+    if (/[\p{Uppercase_Letter}\s]/u.test(text.contents)) {
+      const location = offsetToLineAndColumn(algorithmSource, text.location.start.offset);
+      report({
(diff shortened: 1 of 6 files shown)


src/lint/rules/enum-casing.ts
```diff
@@ -0,0 +1,25 @@
...
+/*
+Checks that ~enum-values~ are kebab-cased.
+*/
+export default function (report: Reporter, step: OrderedListItemNode, algorithmSource: string) {
+  for (const item of step.contents) {
+    if (item.name !== 'tilde' || item.contents.length !== 1 || item.contents[0].name !== 'text') {
+      continue;
+    }
+    const text = item.contents[0];
+    if (/[A-Z ]/.test(text.contents)) {
```
michaelficarra, line 16:
```suggestion
    if (!/^[\p{Ll}\d]+(?:-[\p{Ll}\d]+)*$/v.test(text.contents)) {
```

src/lint/rules/enum-casing.ts
```diff
@@ -0,0 +1,25 @@
...
+      const location = offsetToLineAndColumn(algorithmSource, text.location.start.offset);
+      report({
+        ruleId,
+        message: 'enum values should be kebab-cased',
```
michaelficarra, line 20:
```suggestion
        message: 'enum values should be lowercase and kebab-cased',
```

michaelficarra (commented):
LGTM otherwise

src/lint/rules/enum-casing.ts
bakkot: Wait what? Doesn't kebab-case imply lowercase?

michaelficarra: `THIS-IS-STILL-KEBAB`

michaelficarra (approved):

PR #552, opened 2023-08-31
add a persistent indicator to link targets

michaelficarra:
Split from #441, which only gives a temporary highlight to link targets. This is a further enhancement to add a persistent indicator.

### DIFF
--- a/css/elements.css
+++ b/css/elements.css
@@ -769,6 +769,16 @@ table.lightweight-table th {
   vertical-align: baseline;
 }
 
+/* add a more subtle persistent indicator to link targets */
+emu-clause:target, emu-annex:target, emu-table:target figure, emu-production:target {
+  border-left: 0.5em solid #FFC107;
+  padding-left: 1em;
+  margin-left: -1em;
+}
+:not(emu-clause, emu-annex, emu-table, emu-production):target {
+  text-decoration: wavy underline #FF9800;
+}
+
 /* diff styles */
 ins {
   background-color: #e0f8e0;


css/elements.css
```diff
@@ -769,6 +769,17 @@ table.lightweight-table th {
   vertical-align: baseline;
 }
 
+/* add a more subtle persistent indicator to link targets */
+emu-clause:target, emu-annex:target, emu-table:target figure, emu-production:target {
+  border-left: 0.5em solid #FFC107;
+  padding-left: 1em;
+  margin-left: -1em;
+}
```
gibson042, line 777: This makes the page content jump around... I think it would be better to take advantage of the visual gutter:
```suggestion
/* add a more subtle persistent indicator to link targets */
emu-clause, emu-annex, emu-table figure, emu-production {
  position: relative;
}
emu-clause:target::before, emu-annex:target::before, emu-table:target figure::before, emu-production:target::before {
  position: absolute;
  top: 0;
  height: 100%;
  /* occupy some of #spec-container's 20px horizontal padding */
  left: -18px;
  width: 9px;
  background: #FFC107;
  content: '';
}
```


But also note that making sections `position: relative` necessitates generalizing Toolbox activation code in [js/menu.js](https://github.com/tc39/ecmarkup/blob/b9e720f651c59157fe1fdbe4073cd993d2159661/js/menu.js):
```patch
diff --git i/js/menu.js w/js/menu.js
index adffa64..b2f1d49 100644
--- i/js/menu.js
+++ w/js/menu.js
@@ -914,8 +914,8 @@ let Toolbox = {
     this.entry = entry;
     this.$pinLink.textContent = menu._pinnedIds[entry.id] ? 'Unpin' : 'Pin';
     this.$outer.classList.add('active');
-    this.top = el.offsetTop - this.$outer.offsetHeight;
-    this.left = el.offsetLeft - 10;
+    this.top = getAbsoluteOffset(el, 'offsetTop') - this.$outer.offsetHeight;
+    this.left = getAbsoluteOffset(el, 'offsetLeft') - 10;
     this.$outer.setAttribute('style', 'left: ' + this.left + 'px; top: ' + this.top + 'px');
     this.updatePermalink();
     this.updateReferences();
```
where `getAbsoluteOffset` is something like
```js
function getAbsoluteOffset(el, offsetName) {
  // "offsetTop" => "borderTopWidth", "offsetLeft" => "borderLeftWidth"
  const borderWidthName = offsetName.replace(/offset(.*)/, 'border$1Width');
  let d = el[offsetName];
  for (let parent = el.offsetParent; parent; parent = parent.offsetParent) {
    const parentStyle = getComputedStyle(parent);
    const parentBorderWidth = parentStyle[borderWidthName];
    const parentBorderPx = parseFloat(parentBorderWidth);
    d += (parentBorderPx || 0) + parent[offsetName];
  }
  return d;
}
```

gibson042 (approved):
I like the visuals, just a suggestion on how to better achieve them.

PR #553, opened 2023-09-01
Generalize toolbox positioning to support relative-offset context elements

gibson042:
cf. https://github.com/tc39/ecmarkup/pull/552/files#r1312124976


michaelficarra:
Let's hold off on this until we get approval on the strategy for #552.

gibson042:
Why? Assuming static positioning for so many elements is unnecessarily brittle.

michaelficarra:
I just don't want to introduce the complexity until it buys us something.

gibson042:
Well, I'd like to adopt #552 with a local stylesheet but doing so breaks the toolbox positioning without something like this. I can address that locally as well, but the code here is strictly a generalization—when `el.offsetParent` itself has no offsetParent (as is the case for relevant elements with the current stylesheet), then `getAbsoluteOffset(el, 'offsetTop')` is equivalent to the `el.offsetTop` that it's replacing, and likewise for offsetLeft. It only has an effect when that assumption doesn't hold, in which case the effect is to fix what would otherwise be a bug.

PR #554, opened 2023-09-04
Separate toolbox items into independently-wrappable "words"

gibson042:
The current toolbox cannot wrap in between "Permalink" or "Pin" or "Pin" and "References (…)" (and thus can horizontally overflow the viewport when a definition is near the right margin), but _can_ wrap in between "References" and "(…)" (splitting a link and on the second line intruding into what otherwise looks like box padding). This PR disables wrapping inside those links and inserts break opportunities in between them.

<details><summary><b>Before</b></summary>

screenshot (before change)

</details>

<details><summary><b>After</b></summary>

screenshot (after change)

</details>

The excess horizontal width is a known issue with CSS "shrink-wrapping" that already exists and is not changed by this PR:

<details><summary>current excessive width</summary>

screenshot of current excessive width

</details>

### DIFF
--- a/css/elements.css
+++ b/css/elements.css
@@ -1204,13 +1204,14 @@ li.menu-search-result-term:before {
   background: #ddd;
   border: 1px solid #aaa;
   color: #eee;
-  padding: 5px;
+  padding: 5px 7px;
   border-radius: 3px;
 }
 
 .toolbox a {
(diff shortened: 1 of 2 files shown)


bakkot (approved):

bakkot:
Thanks!

PR #558, opened 2023-11-01
Better completion regex

bakkot:
Regex change is necessary for https://github.com/tc39/ecma262/pull/3163 to land.

The change to the regexp is
- to make it so that only "throw a" as in "throw a TypeError exception" counts, so that a reference to a "throw completion" isn't interpreted as throwing,
- to make "Return Completion Record { ... }" count as returning a completion record, and
- to distinguish between a reference to a "return completion" and a step which says to "Return Completion (...)".

### DIFF
--- a/src/Algorithm.ts
+++ b/src/Algorithm.ts
@@ -92,8 +92,7 @@ export default class Algorithm extends Builder {
               if (completionyThing != null) {
                 if (returnType?.kind === 'completion') {
                   containsAnyCompletionyThings = true;
-                } else if (clause.aoid !== 'GeneratorStart') {
-                  // TODO: remove above exception when the spec is more coherent (https://github.com/tc39/ecma262/pull/2429)
+                } else {
                   spec.warn({
                     type: 'contents',
                     ruleId: 'completiony-thing-in-non-completion-algorithm',
--- a/src/Algorithm.ts
+++ b/src/Algorithm.ts
@@ -87,7 +87,7 @@ export default class Algorithm extends Builder {
                 continue;
               }
               const completionyThing = part.contents.match(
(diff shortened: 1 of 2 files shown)


src/Algorithm.ts
```diff
@@ -87,13 +87,12 @@ export default class Algorithm extends Builder {
                 continue;
               }
               const completionyThing = part.contents.match(
-                /\b(ReturnIfAbrupt|throw|Return (Normal|Throw)?Completion|the result of evaluating)\b|(?<=[\s(])\?\s/i
+                /\b(ReturnIfAbrupt|(^|(?<=, ))[tT]hrow a|[rR]eturn (Normal|Throw|Return)?Completion\(|[rR]eturn( a| a new| the)? Completion Record|the result of evaluating)(\b|$)|(?<=[\s(])\?\s/
```
michaelficarra, line 90:
```suggestion
                /\b(ReturnIfAbrupt\b|(^|(?<=, ))[tT]hrow (a\b|the\b|_|\*)|[rR]eturn (Normal|Throw|Return)?Completion\(|[rR]eturn( a| a new| the)?Completion Record\b|the result of evaluating\b)|(?<=[\s(])\?\s/
```

This will permit "throw the", "throw \_alias\_", "throw \*true\*", etc.

bakkot: This is running on the parsed ecmarkdown, not the plain text input, so matching `_` and `*` would require a much larger refactor which I don't really want to do. I could add `the` I guess? But in practice that hasn't come up.

bakkot: Though I guess actually `throw $` would match text followed by `*`, `~`, `_`, etc, which is probably good enough. Updated.

michaelficarra (approved):

PR #560, opened 2023-11-02
add unpinning buttons and clear button to pin list

### LINKED ISSUE #438: the pinned items should have a button next to them for easy unpinning
It would be nice if the pinned items had a button next to them for easy unpinning, and an "unpin all" feature would be even nicer.

_Originally posted by @michaelficarra in https://github.com/tc39/ecmarkup/issues/399#issuecomment-996357637_

michaelficarra:
Fixes #438.

image


css/elements.css
```diff
@@ -1285,12 +1326,16 @@ li.menu-search-result-term:before {
...
+  cursor: row-resize;
+}
+
 #references-pane-container.active {
   display: flex;
 }
 
 #references-pane-close:after {
-  content: '✖';
+  content: '\2716';
```
michaelficarra, line 1338: This is a drive-by fix for an encoding error I got when doing local testing from a `file://` URL.

bakkot: ... is that seriously how one spells hex escapes in CSS? Gross.

michaelficarra: yep https://drafts.csswg.org/css-syntax/#escaping

bakkot (approved):
Very nice.

PR #561, opened 2023-11-02
add support for keyboard 0 selecting pin number 10

michaelficarra:
Currently only 1-9 select pins 1-9. Adds support for 0 selecting pin 10.


michaelficarra:
Just to be certain, I checked https://en.wikipedia.org/wiki/Keyboard_layout and it seems like every keyboard layout in use around the world orders the numerals 1-9 followed by 0.

bakkot (approved):
Gross but I'll take it.

PR #564, opened 2023-12-03
Accept dot inside intrinsic name in header

arai-a:
The current rule doesn't accept a dot inside the intrinsic name in header, but there are multiple intrinsics that has dots inside their names, such as `%GeneratorFunction.prototype.prototype%`.

This allows the solution (A) in https://github.com/tc39/ecma262/issues/3230

### DIFF
--- a/src/lint/collect-header-diagnostics.ts
+++ b/src/lint/collect-header-diagnostics.ts
@@ -49,11 +49,17 @@ export function collectHeaderDiagnostics(
 
       // CreateForInIterator
       // Object.fromEntries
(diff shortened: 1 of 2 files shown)


ljharb:
If that solution ends up not being adopted (I personally don't think there's a problem that needs solving, and I think the solution is less readable), would this change be needed?

arai-a:
Closing given https://github.com/tc39/ecma262/pull/3234 is superseded by https://github.com/tc39/ecma262/pull/3238 with %GeneratorPrototype%, which doesn't have a dot inside %.

PR #568, opened 2024-01-07
basic typechecking

bakkot:
Catches errors like https://github.com/tc39/proposal-joint-iteration/pull/11#discussion_r1442155330 via a very basic type system.

The checks only apply to call arguments, and specifically only to arguments which are either literals or themselves calls (including `! Foo()`-style unwrapped calls). There's no tracking of variables or anything yet.

Also, we can't generally infer types precisely. For example, imagine `Identity(x)` which returns its input and accepts any value: `TakesInteger(Identity(42))` would be a reasonable thing to write, but the type system here is much to weak to know that. So we check only that the intersection (that is, the meet) of the argument type and parameter type is nonempty.

The first commit is just moving code around. I recommend reviewing commits individually.

### DIFF
--- a/src/typechecker.ts
+++ b/src/typechecker.ts
@@ -600,8 +600,7 @@ function serialize(type: Type): string {
     case 'integer':
     case 'non-negative integer':
     case 'null':
-    case 'undefined':
-    case 'string': {
+    case 'undefined': {
       return type.kind;
     }
     case 'concrete string':
@@ -623,6 +622,9 @@ function serialize(type: Type): string {
     }
     case 'boolean': {
       return 'Boolean';
+    }
+    case 'string': {
+      return 'String';
     }
     case 'number': {
       return 'Number';
--- a/src/typechecker.ts
+++ b/src/typechecker.ts
@@ -386,6 +386,8 @@ type Type =
   | { kind: 'real' }
   | { kind: 'integer' }
   | { kind: 'non-negative integer' }
+  | { kind: 'negative integer' }
+  | { kind: 'positive integer' }
   | { kind: 'concrete real'; value: string }
   | { kind: 'ES value' }
   | { kind: 'string' }
@@ -410,6 +412,8 @@ const simpleKinds = new Set<Type['kind']>([
   'real',
   'integer',
   'non-negative integer',
+  'negative integer',
+  'positive integer',
   'ES value',
   'string',
   'number',
@@ -424,8 +428,9 @@ const dominateGraph: Partial<Record<Type['kind'], Type['kind'][]>> = {
   // @ts-expect-error TS does not know about __proto__
   __proto__: null,
   record: ['completion'],
-  real: ['integer', 'non-negative integer', 'concrete real'],
-  integer: ['non-negative integer'],
+  real: ['integer', 'non-negative integer', 'negative integer', 'positive integer', 'concrete real'],
+  integer: ['non-negative integer', 'negative integer', 'positive integer'],
+  'non-negative integer': ['positive integer'],
   'ES value': [
     'string',
     'number',
@@ -484,11 +489,18 @@ function dominates(a: Type, b: Type): boolean {
   if (a.kind === 'non-negative integer' && b.kind === 'concrete real') {
     return !b.value.includes('.') && b.value[0] !== '-';
   }
+  if (a.kind === 'negative integer' && b.kind === 'concrete real') {
+    return !b.value.includes('.') && b.value[0] === '-';
+  }
+  if (a.kind === 'positive integer' && b.kind === 'concrete real') {
+    return !b.value.includes('.') && b.value[0] !== '-' && b.value !== '0';
+  }
   if (
     a.kind === b.kind &&
     [
       'concrete string',
       'concrete number',
+      'concrete real',
       'concrete bigint',
       'concrete boolean',
       'enum value',
@@ -599,6 +611,8 @@ function serialize(type: Type): string {
     }
     case 'integer':
     case 'non-negative integer':
+    case 'negative integer':
+    case 'positive integer':
     case 'null':
     case 'undefined': {
       return type.kind;
@@ -763,6 +777,9 @@ function typeFromExprType(type: BiblioType): Type {
       if (text.startsWith('~') && text.endsWith('~')) {
         return { kind: 'enum value', value: text.slice(1, -1) };
       }
+      if (/^-?[0-9]+(\.[0-9]+)?$/.test(text)) {
+        return { kind: 'concrete real', value: text };
+      }
       if (text === 'an ECMAScript language value' || text === 'ECMAScript language values') {
         return { kind: 'ES value' };
       }
@@ -789,6 +806,12 @@ function typeFromExprType(type: BiblioType): Type {
       }
       if (text === 'a non-negative integer' || text === 'non-negative integers') {
         return { kind: 'non-negative integer' };
+      }
+      if (text === 'a negative integer' || text === 'negative integers') {
+        return { kind: 'negative integer' };
+      }
+      if (text === 'a positive integer' || text === 'positive integers') {
+        return { kind: 'positive integer' };
       }
       if (text === '*null*') {
         return { kind: 'null' };
--- a/src/typechecker.ts
+++ b/src/typechecker.ts
@@ -762,7 +762,7 @@ function typeFromExprType(type: BiblioType): Type {
       if (text === 'an ECMAScript language value' || text === 'ECMAScript language values') {
         return { kind: 'ES value' };
       }
-      if (text === 'a string' || text === 'strings') {
+      if (text === 'a String' || text === 'Strings') {
         return { kind: 'string' };
       }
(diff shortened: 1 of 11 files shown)


src/typechecker.ts
```diff
@@ -298,3 +357,397 @@ function textFromPreviousPart(seq: Seq, index: number): string | null {
...
+    return true;
+  }
+  if (dominateGraph[a.kind]?.includes(b.kind) ?? false) {
+    return true;
+  }
+  if (
+    (a.kind === 'integer' && b.kind === 'concrete real') ||
+    (a.kind === 'integral number' && b.kind === 'concrete number')
+  ) {
+    return !b.value.includes('.');
```
michaelficarra, line 468: `1.23456789e+23` is integral and a canonical ToString. 

```suggestion
    let n = parseFloat(b.value);
    return n === Math.round(n);
```

Alternatively, you could just keep a number instead of a string.

src/typechecker.ts
```diff
@@ -298,3 +357,397 @@ function textFromPreviousPart(seq: Seq, index: number): string | null {
...
+  }
+  if (a.kind === 'non-negative integer' && b.kind === 'concrete real') {
+    return !b.value.includes('.') && b.value[0] !== '-';
+  }
+  if (
+    a.kind === b.kind &&
+    ['concrete string', 'concrete number', 'concrete bigint', 'enum value'].includes(a.kind)
+  ) {
+    // @ts-expect-error TS is not quite smart enough for this
```
michaelficarra, line 477: There's a couple ways that you could avoid this. Either use some predicates that narrow `a` or a big disjunction.

bakkot: Yeah but I don't want to write more complicated code just to please the type checker, since this code is in fact clearly correct. That's what `@ts-expect-error` is for.

src/typechecker.ts
bakkot: Do we actually want to allow people to write numbers like `*1.23456789e+23*<sub>𝔽</sub>`?

src/typechecker.ts
```diff
@@ -298,3 +357,397 @@ function textFromPreviousPart(seq: Seq, index: number): string | null {
...
+
+      const biblioEntry = biblio.byAoid(calleeName);
+      if (biblioEntry?.signature?.return == null) {
+        break;
+      }
+      return typeFromExprType(biblioEntry.signature.return);
+    }
+    case 'tilde': {
+      if (expr.contents.length === 1 && expr.contents[0].name === 'text') {
+        return { kind: 'enum value', value: `~${expr.contents[0].contents}~` };
```
michaelficarra, line 676: I wouldn't bother including the tildes.

src/typechecker.ts
michaelficarra: I guess I would slightly lean toward only permitting `DecimalDigits` and `DecimalDigits . DecimalDigits` and requiring that they be an exactly-representable float value. This is an editorial question we should confirm at an editor call.

Regardless though, I would still change the representation to `number` and move the complexity of what's permitted upstream.

test/typecheck.js
```diff
@@ -1174,3 +1174,189 @@ describe('negation', async () => {
...
+
+describe('type system', () => {
+  async function assertTypeError(paramType, arg, message, extraBiblios = []) {
+    await assertLint(
+      positioned`
+        <emu-clause id="example" type="abstract operation">
+          <h1>
+            Example (
+              arg: ${paramType}
+            )
```
michaelficarra, line 1186: `): ~unused~`, right?

test/typecheck.js
```diff
@@ -1174,3 +1174,189 @@ describe('negation', async () => {
...
+  }
+
+  it('enum', async () => {
+    await assertTypeError(
+      '~sync~ or ~async~',
+      '~iterate-strings~',
+      'argument (~iterate-strings~) does not look plausibly assignable to parameter type (~sync~ or ~async~)'
+    );
+
+    await assertNoTypeError('~sync~ or ~async~', '~sync~');
```
michaelficarra, line 1231: also `await assertNoTypeError('~sync~ or ~async~', '~async~');` please

src/typechecker.ts
bakkot: Changed.

bakkot:
> You should add support for undefined/booleans.

Done.

michaelficarra:
It looks like you're missing some coverage in the tests. For example, I don't see anything for EcmaScript language values. I would expect the whole dominator lattice to be tested.

test/typecheck.js
bakkot: Isn't required and doesn't affect anything, but sure.

src/typechecker.ts
```diff
@@ -0,0 +1,788 @@
...
+      if (text === 'a non-negative integer' || text === 'non-negative integers') {
+        return { kind: 'non-negative integer' };
+      }
+      if (text === '*null*') {
+        return { kind: 'null' };
+      }
+      if (text === '*undefined*') {
+        return { kind: 'undefined' };
+      }
+      break;
```
michaelficarra, line 780: Instead of returning `{ kind: 'unknown' }` here, why not have an `opaque` kind that just compares textually? Like if I have an AO that returns "a Foo" and another AO that accepts "a Bar", that should be rejected.

bakkot: Specifically because that should _not_ be rejected. Maybe all Foos are Bars. We don't have enough information to say definitively that that's wrong.

test/typecheck.js
```diff
@@ -1174,3 +1174,325 @@ describe('negation', async () => {
...
+  }
+
+  async function assertNoTypeError(paramType, arg, extraBiblios = []) {
+    await assertLintFree(
+      `
+        <emu-clause id="example" type="abstract operation">
+          <h1>
+            Example (
+              arg: ${paramType}
+            )
```
michaelficarra, line 1212: Also here.

```suggestion
            ): ~unused~
```

src/typechecker.ts
michaelficarra: Okay. Let's make sure there's a positive test for it then.

src/typechecker.ts
michaelficarra: 1 is exactly representable though...

bakkot: Pretend I said `0.1` like I obviously meant.

michaelficarra: That could be fine if we add some explanation in the Notational Conventions section saying that we mean the nearest-representable float. This is a topic for editor call.

src/typechecker.ts
michaelficarra: We could also probably use a mode that operates like this to find other useful low-hanging fruit we could support.

src/typechecker.ts
bakkot: Also, changing this to be a `number` doesn't work for reals, which can in principle be values which are not representable as doubles. So I would still need to keep the string-handling logic for the real case. Given that, I don't want to introduce a different path just four doubles. That's a lot of extra complexity for no benefit.

test/typecheck.js
bakkot: Done.

src/typechecker.ts
michaelficarra: You already have different paths for reals and doubles.

bakkot: In some places, but not in this case.

src/typechecker.ts
michaelficarra: Also, `a BigInt` should be the type for bigints, right?

bakkot: Those are _parameter_ types, rather than expressions; there's a different path for converting types in signatures into type-system types than there is for converting expressions in algorithms into type-system types.

michaelficarra (approved):

src/typechecker.ts
```diff
@@ -0,0 +1,896 @@
...
+      if (Object.is(type.value, -0)) {
+        repr = '-0';
+      } else if (type.value === 0) {
+        repr = '+0';
+      } else if (type.value === 2e308) {
+        repr = '+&infin;';
+      } else if (type.value === -2e308) {
+        repr = '-&infin;';
+      } else if (type.value > 4503599627370495.5) {
+        repr = String(BigInt(type.value));
```
bakkot, line 652: This will throw if you pass an excessively large float. The parser shouldn't throw. It can surface an error to the user, but it shouldn't throw.

michaelficarra: Will it? What's an example?

```
> String(BigInt(1e308))
< '100000000000000001097906362944045541740492309677311846336810682903157585404911491537163328978494688899061249669721172515611590283743140088328307009198146046031271664502933027185697489699588559043338384466165001178426897626212945177628091195786707458122783970171784415105291802893207873272974885715430223118336'
```

michaelficarra: This is the largest non-integral double.

bakkot: Oh, fair enough. WFM then.

PR #571, opened 2024-01-31
reduce document width on wider displays

michaelficarra:
before:

image

after:

image

We could reduce it more but I figured I'd start with a smaller change.

/cc @devsnek

### DIFF
--- a/css/elements.css
+++ b/css/elements.css
@@ -110,11 +110,17 @@ body {
 
 #spec-container {
   padding: 0 20px;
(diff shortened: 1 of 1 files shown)


bakkot:
This seems way worse to me.

Can we at least go with the HTML spec's 80em instead of 50?

michaelficarra:
Updated to 80rem

image

bakkot (approved):

Issue #572, opened 2024-02-20
Allow rest argument list in abstract closures


legendecas:
ecmarkup complains about a rest arguments list in an abstract closure declaration, such as:

```html
<emu-alg>
  1. If IsCallable(_fn_) is false, throw a TypeError exception.
  1. Let _closure_ be a new Abstract Closure with parameters (..._args_) that captures _fn_ and performs the following steps when called:
    1. Let _thisArgument_ be the *this* value.
    1. Let _result_ be Completion(Call(_fn_, _thisArgument_, _args_)).
    1. Return _result_.
  1. Let _length_ be ? LengthOfArrayLike(_fn_).
  1. Let _name_ be ? Get(_fn_, *"name"*).
  1. If _name_ is not a String, set _name_ to the empty String.
  1. Return CreateBuiltinFunction(_closure_, _length_, _name_, &laquo; &raquo;).
</emu-alg>
```

Run with command `ecmarkup --verbose --lint-spec spec.html build/index.html`, and it outputs errors:

```
[2024-02-20T16:42:00.406Z] Error: spec.html:903:71: expected to find a parameter name here
[2024-02-20T16:42:01.873Z] Completed with errors.
error: expected to find a parameter name here (bad-ac) at spec.html:903:71:
> 903 |           1. Let _closure_ be a new Abstract Closure with parameters (..._args_) that captures _fn_ and performs the following steps when called:
```

This pattern can be handy compared to the exotic function object language.

Refs: https://github.com/tc39/proposal-async-context/pull/68#discussion_r1493400361

PR #573, opened 2024-02-23
Implement rest params in abstract closures

jridgewell:
Fixes https://github.com/tc39/ecmarkup/issues/572


bakkot (approved):
Thanks! Sorry you had to touch this snippet; that bit of parsing is, shall we say, even more ad-hoc than the rest of it.

bakkot:
Published in v18.2.0.

PR #574, opened 2024-03-07
support a separate print stylesheet

michaelficarra:
Print styles live in `css/print.css` now.

### DIFF
--- a/src/Spec.ts
+++ b/src/Spec.ts
@@ -1035,7 +1035,7 @@ ${await utils.readFile(path.join(__dirname, '../js/multipage.js'))}
     }
 
     let cssContents = await utils.readFile(path.join(__dirname, '../css/elements.css'));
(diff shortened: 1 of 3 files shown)


src/Spec.ts
```diff
@@ -1108,10 +1115,13 @@ ${await utils.readFile(path.join(__dirname, '../js/multipage.js'))}
     );
   }
 
-  private addStyle(head: HTMLHeadElement, href: string) {
+  private addStyle(head: HTMLHeadElement, href: string, print: boolean = false) {
```
bakkot, line 1118: Maybe
```suggestion
  private addStyle(head: HTMLHeadElement, href: string, media?: string) {
```
for the sake of avoiding boolean attributes?

michaelficarra: The only allowed values are `all`/`print`/`screen` but okay fine.

bakkot (approved):

PR #575, opened 2024-03-08
implement some basic print styles

michaelficarra:
These follow Ecma's guidelines.


css/print.css
```diff
@@ -1,3 +1,28 @@
-#menu-toggle {
-  display: none;
```
bakkot, line 2: Why get rid of this?

michaelficarra: that element wasn't present when using the old TOC

bakkot: It is not yet the case that "printing" is equivalent to "old TOC".

I guess it's fine landing now but I'd've preferred doing it as part of that.

bakkot: I guess the new TOC is hidden when printing anyway so it doesn't matter.

bakkot (approved):

PR #579, opened 2024-03-09
add page numbers after ToC entries

michaelficarra:


bakkot (approved):
Neat!

Would be nice to generate a glossary using this as well, someday.

michaelficarra:
Okay this is pretty neat.

image

michaelficarra:
Dotted leaders are working now:

image

src/Toc.ts
```diff
@@ -26,7 +26,14 @@ export default class Toc {
     this.spec.doc.body.setAttribute('class', bodyClass + ' oldtoc');
   }
 
-  static build(level: Spec | Clause, expandy?: boolean) {
+  static build(level: Spec | Clause, options: { maxDepth?: number, expandy?: boolean } = {}) {
```
bakkot, line 29: If you destructure in place you can set the defaults inline and skip the explicit type:
```suggestion
  static build(level: Spec | Clause, { maxDepth = Infinity, expandy = false } = {}) {
```

src/Toc.ts
```diff
@@ -45,7 +52,7 @@ export default class Toc {
         html += '<span class="secnum">' + sub.number + '</span> ';
       }
       html += shorten(sub.titleHTML) + '</a>';
-      if (sub.subclauses.length > 0) html += Toc.build(sub, expandy);
+      if (sub.subclauses.length > 0) html += Toc.build(sub, Object.assign({}, options, { maxDepth: maxDepth - 1 }));
```
bakkot, line 55:
```suggestion
      if (sub.subclauses.length > 0) html += Toc.build(sub, {...options, maxDepth: maxDepth - 1 });
```

bakkot (approved):

PR #580, opened 2024-03-19
handle image assets similarly to how we handle font asset inlining

michaelficarra:


michaelficarra:
Also some changes came in through `npm run format`.

bakkot (approved):

PR #581, opened 2024-03-20
print CSS improvements

michaelficarra:


css/print.css
```diff
@@ -11,6 +11,38 @@ body {
 
 p {
   text-align: justify;
+  text-rendering: optimizeLegibility;
```
bakkot, line 14: MDN says this only works on Linux/Windows; something to watch out for.

bakkot (approved):

PR #588, opened 2024-03-22
avoid iterating a live nodelist

bakkot:
This step is still surprisingly slow but this makes it take 5 seconds instead of 10 on my machine.

### DIFF
--- a/src/Spec.ts
+++ b/src/Spec.ts
@@ -819,7 +819,7 @@ export default class Spec {
   }
 
   private annotateExternalLinks(): void {
-    for (const a of this.doc.getElementsByTagName('a')) {
+    for (const a of [...this.doc.getElementsByTagName('a')]) {
       if (a.hostname !== '' && a.href !== a.textContent && a.protocol !== 'mailto:') {
         a.setAttribute('data-print-href', '');
       }
--- a/src/Spec.ts
+++ b/src/Spec.ts
@@ -819,7 +819,7 @@ export default class Spec {
   }
 
   private annotateExternalLinks(): void {
-    for (const a of [...this.doc.getElementsByTagName('a')]) {
+    for (const a of this.doc.querySelectorAll('a')) {
       if (a.hostname !== '' && a.href !== a.textContent && a.protocol !== 'mailto:') {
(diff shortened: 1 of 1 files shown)


michaelficarra (approved):

gibson042:
It occurs to me that even more time could be saved by skipping this step entirely and instead updating the print CSS to use relevant attribute-value-starts-with selectors:
```diff
+/* Render non-mailto external link targets, cf. https://url.spec.whatwg.org/#special-scheme */
-a[data-print-href]::after {
+a:is([href^='ftp:' i], [href^='sftp:' i], [href^='http:' i], [href^='https:' i], [href^='ws:' i], [href^='wss:' i])::after {
   content: ' <' attr(href) '>';
   color: initial;
 }
```

This would miss new schemes until added to the CSS (_which is lintable_) and would be redundant for links whose text matches their target (_which we should probably not have anyway—it's more accessible to name links and let user agents handle URIs_).

ljharb:
also if you're going for speed, you should avoid `for..of` aggressively.

michaelficarra:
> and would be redundant for links whose text matches their target (_which we should probably not have anyway—it's more accessible to name links and let user agents handle URIs_)

Yes but we do and that's why we can't use this CSS-only strategy.

> also if you're going for speed, you should avoid `for..of` aggressively.

The for-of is only iterating 94 elements. It's not the slow part. The slow part, for some unexplainable reason, is setting an attribute.

bakkot:
> The for-of is only iterating 94 elements.

(Well, no, it's iterating 30k elements, but it's still true that the iteration itself is now ~instant.)

gibson042:
> > and would be redundant for links whose text matches their target (_which we should probably not have anyway—it's more accessible to name links and let user agents handle URIs_)
> 
> Yes but we do and that's why we can't use this CSS-only strategy.

Having them doesn't preclude a CSS-only approach, it just means that such an approach would duplicate URLs. But to address the larger point, would you object to fixing such links by giving them non-URL text?

gibson042:
ecma262 PR: https://github.com/tc39/ecma262/pull/3301

Issue #416, opened 2022-03-09
Published `.d.ts` files don't type-check


DanielRosenwasser:
If you try to consume the `ecmarkup` package via the API, it will fail to type-check because `Clause` is marked `/* @internal */`...

https://github.com/tc39/ecmarkup/blob/05d05f71e1bc8cf151da65d34e9105eff3e7d465/src/Clause.ts#L26-L27

but `maybeAddClauseToEffectWorklist` still uses the non-existent default export publicly.

https://github.com/tc39/ecmarkup/blob/05d05f71e1bc8cf151da65d34e9105eff3e7d465/src/Spec.ts#L254-L258

This results in the error:

```
node_modules/ecmarkup/lib/Spec.d.ts:49:84 - error TS2709: Cannot use namespace 'Clause' as a type.
```

Should `maybeAddClauseToEffectWorklist` be marked internal as well?

bakkot:
@DanielRosenwasser I can remove `/* @internal */` from `Clause`, but ideally I'd like to figure out how to prevent this category of issue in the future - is there a way I can write a test which ensures the published TS definitions typecheck? I didn't realize TS would let me publish invalid `.d.ts` files from a well-typed project, given that it's the thing generating them in the first place...

rbuckton:
`/* @internal */` is a long-standing "undocumented" feature that I've wanted to redo [since 2015](https://github.com/microsoft/TypeScript/issues/5228) into something actually baked into the language. It was a bit of an expert-level workaround for trimming things from your public API and unfortunately hasn't received much attention since TS 1.0.

PR #591, opened 2024-03-28
Simple test to validate declaration emit in presence of `@internal`

rbuckton:
Ahead of actually fixing issues with the declarations, this adds a simple test that can be manually invoked to validate declaration emit in the presence of `/* @internal */` and `--stripInternal`.

Long term, I would recommend refactoring out internals such that `/* @internal */` is unnecessary.

Related: https://github.com/tc39/ecmarkup/issues/416#issuecomment-1074739611

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -33,3 +33,6 @@ jobs:
 
       - name: Test published files
         run: npm run test-published-files
+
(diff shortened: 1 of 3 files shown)


bakkot:
Merged as part of https://github.com/tc39/ecmarkup/pull/592.

PR #592, opened 2024-03-28
Move `@internal` markers to individual members

### LINKED ISSUE #416: Published `.d.ts` files don't type-check

rbuckton:
This builds on #591 by moving `/*@internal*/` markers from classes to individual class elements so that class types are preserved broadly while still allowing us to remove individual `/*@internal*/` markers on a case-by-case basis.

Fixes #416

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -33,3 +33,6 @@ jobs:
 
       - name: Test published files
         run: npm run test-published-files
+
+      - name: Test declarations
+        run: npm run test-declarations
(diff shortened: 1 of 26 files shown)


bakkot (approved):
Sweet, thank you!

PR #595, opened 2024-05-23
update lint rule to accept %-notation in operation headers

michaelficarra:
Fixes CI failure for https://github.com/tc39/ecma262/pull/1314. /cc @ljharb


ljharb (approved):
LGTM! rejecting the @@ notation makes it a breaking change (nbd, ofc, just calling it out)

bakkot (approved):

PR #596, opened 2024-06-07
Make the formatter smarter about mutli-line steps

bakkot:
Some consumers write some steps across multiple line of source, [like this](https://github.com/tc39/proposal-temporal/blob/26e4cebe3c49f56932c1d5064fec9993e981823a/spec/abstractops.html#L621-L625). This makes the formatter smarter in those cases: rather than making all lines be the same indentation, it keeps whatever indentation there is beyond that which is common across all lines.

I would have been happier making it aware of Records explicitly, but that's quite complex: the [expression parser](https://github.com/tc39/ecmarkup/blob/fd938c8a4da2dca309f8c54a72d8c8616f5da009/src/expr-parser.ts#L737-L738) intentionally doesn't represent `<del>` etc, which means its output is not suitable for use in the formatter.

### DIFF
--- a/src/expr-parser.ts
+++ b/src/expr-parser.ts
@@ -10,7 +10,7 @@ type SimpleLocation = { start: { offset: number }; end: { offset: number } };
 type BareText = {
   name: 'text';
   contents: string;
-  location: { start: { offset: number }; end: { offset: number } };
+  location: SimpleLocation;
 }; // like TextNode, but with less complete location information
 type ProsePart = FragmentNode | BareText;
 type List = {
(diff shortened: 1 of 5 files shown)


michaelficarra (approved):
I'm gonna take your word for it.

gibson042:
This seems related to #513, but I'm not sure to what extent.

bakkot:
@gibson042 This is only about the text representation, not the output. Though of course you need the text representation to be preserved by the formatter if you're going to use it to guide the output.

Issue #593, opened 2024-04-10
include `meta property="og:description"` tags


bakkot:
so social media previews don't look like this:

<img width="1025" alt="Screenshot 2024-04-10 at 9 04 39 AM" src="https://github.com/tc39/ecmarkup/assets/1653598/eba784cc-3472-4ee4-96c0-1240a3b7b883">

The user can supply one in the metadata block. We should remember to update the template-for-proposals so people actually remember to do so, also.

PR #598, opened 2024-06-07
add support for opengraph tags

bakkot:
Fixes https://github.com/tc39/ecmarkup/issues/593:

Screenshot 2024-06-06 at 10 11 19 PM

I've hardcoded [the URL of the ECMA logo in this project's docs](https://tc39.es/ecmarkup/ecma-header.png) as the `og:image`, which isn't my favorite option, but it does work.


src/Spec.ts
```diff
@@ -1438,6 +1438,41 @@ ${this.opts.multipage ? `<li><span>Navigate to/from multipage</span><code>m</cod
...
+    const insertMetaTag =
+      metas.length === 0
+        ? (node: HTMLMetaElement) => this.doc.head.prepend(node)
+        : (node: HTMLMetaElement) => metas[metas.length - 1].after(node);
+    if (!metas.some(n => n.getAttribute('property') === 'og:description')) {
+      const description =
+        this.opts.description ??
+        (this.doc.querySelector('emu-intro') ?? this.doc.querySelector('emu-clause'))?.textContent
+          ?.slice(0, 300)
+          .trim();
```
michaelficarra, line 1453:
```suggestion
        (this.opts.description ?? (this.doc.querySelector('emu-intro') ?? this.doc.querySelector('emu-clause'))?.textContent)
          ?.slice(0, 300)
          .trim();
```

Maybe also slice the explicit description?

src/Spec.ts
```diff
@@ -1438,6 +1438,41 @@ ${this.opts.multipage ? `<li><span>Navigate to/from multipage</span><code>m</cod
...
+        const meta = this.doc.createElement('meta');
+        meta.setAttribute('property', 'og:title');
+        meta.setAttribute('content', title);
+        insertMetaTag(meta);
+      }
+    }
+    if (!metas.some(n => n.getAttribute('property') === 'og:image')) {
+      const meta = this.doc.createElement('meta');
+      meta.setAttribute('property', 'og:image');
+      meta.setAttribute('content', 'https://tc39.es/ecmarkup/ecma-header.png');
```
michaelficarra, line 1473: I don't love this choice, but it's better than nothing and we can improve it later.

What about the Ecma logo as a slightly less-bad option?

image

michaelficarra (approved):
LGTM otherwise

src/Spec.ts
bakkot: Happy with that too though it would need to live somewhere static in order to be referenced. I can add it to the docs here.

bakkot: Done. Further improvements of course welcome.

PR #600, opened 2024-06-13
Markup: PDF generation

gesa:
Tried to keep the css together but it doesn't feel super logically grouped so I wound up rearranging everything. Sorry future git-blamer. Units of measurements are all over the place, here's what defines them:

- mm for page-related units
- pt for text-related units
- em for horizontal units
- ex for vertical units

Highly specific rules come from a loyalty to the Ecma specification template first, previous years' standards second.

There should be no changes that impact screen rendering. Any changes to the primary stylesheet come from inline styles that existed in both ECMA-262 and ECMA-402.

The cover SVGs aren't really used by ecmarkup directly, but are rather to be viewed as source files for the two specifications.


css/print.css
```diff
@@ -24,58 +142,274 @@ h1 {
...
+  -prince-bookmark-level: 4;
+}
+
+emu-clause emu-clause emu-clause emu-clause emu-clause h1, emu-annex emu-annex emu-annex emu-annex emu-annex h1 {
+  -prince-bookmark-level: 5;
+}
+
+emu-clause emu-clause emu-clause emu-clause emu-clause emu-clause h1, emu-annex emu-annex emu-annex emu-annex emu-annex emu-annex h1 {
+  -prince-bookmark-level: 6;
+}
```
michaelficarra, line 287: 😭

css/print.css
```diff
@@ -84,119 +418,200 @@ emu-figure img {
...
+  margin-bottom: 1ex;
 }
+#sec-example-cyclic-module-record-graphs emu-figure:has(> figure > img) + p {
+  margin-top: 1ex;
+}
+
+/* 16.2.1.5.4 table 46 is still wide after pivoting */
+#table-module-graph-cycle-async-fields-1 table {
+  font-size: 95%;
+}
```
michaelficarra, line 608: Could you just `colspan` the `[[Status]]` row all the way across so that `~evaluating-async~` doesn't stretch out every column?

css/print.css
```diff
@@ -84,119 +418,200 @@ emu-figure img {
...
+#sec-delete-operator-static-semantics-early-errors emu-grammar {
+  display: inline-block;
 }
 
-@page :left {
-  @bottom-left {
-    content: counter(page, var(--page-number-style));
-  }
+/* 15.3 A very long term combined with inline-block, nowrap, and justified text resulting in weird punctuation */
+#_ref_20642 {
```
michaelficarra, line 582: I don't think this ref will be stable?

css/print.css
```diff
@@ -84,119 +418,200 @@ emu-figure img {
...
 }
-@page :first {
-  @top-center {
-    content: none;
-  }
+
+
+/**
+* ECMA-402-specific hacks
+*/
```
michaelficarra, line 483: I'd prefer if any document-specific hacks live in that document's repo.

src/Spec.ts
```diff
@@ -1214,6 +1198,7 @@ ${await utils.readFile(path.join(__dirname, '../js/multipage.js'))}
     const elements = this.doc.body.childNodes;
     const wrapper = this.doc.createElement('div');
     wrapper.id = 'spec-container';
+    wrapper.setAttribute('data-shortname', this.opts.shortname || '');
```
michaelficarra, line 1201:
```suggestion
    if (this.opts.shortname) {
      wrapper.setAttribute('data-shortname', this.opts.shortname);
    }
```

so we don't have a ton of empty attributes

css/print.css
syg: uwu what's this

gesa: a whole lot of cascading

gesa:
See https://github.com/gesa/ecmarkup/commit/c6b6b9cf6d70c9a2cd9c11941b8236bf1b7b7bbf#comments for the most important detail of this PR.

michaelficarra:
@gesa Have you test both with and without `--multipage`?

gesa:
@michaelficarra I have, I’ve generated web versions of ecma402, ecma262 (multipage), and temporal.

css/elements.css
```diff
@@ -775,6 +777,10 @@ emu-table td {
   background: #fff;
 }
 
+th[colspan], td[colspan] {
```
michaelficarra, line 780:
```suggestion
td[colspan]:not([colspan="1"]), th[colspan]:not([colspan="1"]) {
```

gesa:
@syg , @michaelficarra : following up to get this merged when y’all have a chance

css/elements.css
```diff
@@ -1425,3 +1439,28 @@ li.menu-search-result-term:before {
...
+
+#ecma-logo {
+  width: 500px;
+}
+
+.unicode-property-table {
+  table-layout: fixed;
+  width: 100%;
+  font-size: 80%;
+}
```
michaelficarra, line 1466: I thought we were going to move this into ecma262

gesa: Wait why? That’s not specific to 262.

js/print.js
```diff
@@ -0,0 +1,316 @@
...
+
+  insideCover.classList.add('full-page-svg');
+  insideCover.setAttribute('id', 'inside-cover');
+  insideCover.innerHTML =
+    '<p>Ecma International<br />Rue du Rhone 114 CH-1204 Geneva<br/>Tel: +41 22 849 6000<br/>Fax: +41 22 849 6001<br/>Web: https://www.ecma-international.org<br/>Ecma is the registered trademark of Ecma International.</p>';
+
+  return insideCover;
+}
+
+function generateEcmaCopyrightPage() {
```
syg, line 260: Why are we generating another copyright page? Don't we already have one in the document?

css/elements.css
michaelficarra: Yes, this class only exists on a table in ecma262: https://github.com/tc39/ecma262/blob/815aa8a4d68ce24bec81ba6789c37d55bc87b54a/table-binary-unicode-properties.html#L3.

js/print.js
```diff
@@ -0,0 +1,316 @@
...
+  }
+
+  if (/^[A-Z]\./.test(clauseNumber)) {
+    link.innerHTML = '<span class="secnum">Annex ' + clauseNumber + '</span> ' + clauseTitle;
+    return link;
+  }
+
+  link.innerHTML = '<span class="secnum">' + clauseNumber + '</span> ' + clauseTitle;
+  return link;
+}
```
michaelficarra, line 141: I'm pretty sure we could've done all of this with print styles and an `:after` selector.

js/print.js
```diff
@@ -0,0 +1,316 @@
...
+function removeEmuImports() {
+  const emuImports = Array.from(specContainer.getElementsByTagName('emu-import'));
+
+  emuImports.forEach(importedMarkup => {
+    while (importedMarkup.hasChildNodes()) {
+      importedMarkup.parentNode.insertBefore(importedMarkup.childNodes[0], importedMarkup);
+    }
+    importedMarkup.parentNode.removeChild(importedMarkup);
+  });
+}
```
michaelficarra, line 197: I think it'd be fine to just not emit these in the first place.

js/print.js
```diff
@@ -0,0 +1,316 @@
...
+/**
+ * Gets rid of elements we don't need in the print version
+ * */
+function removeUnusedSections() {
+  const ecmaLogo = document.getElementById('ecma-logo');
+
+  if (ecmaLogo) specContainer.removeChild(ecmaLogo.parentNode);
+
+  document.getElementsByTagName('body')[0].removeChild(document.getElementById('shortcuts-help'));
+}
```
michaelficarra, line 228: Surely this can be done with print CSS.

js/print.js
```diff
@@ -0,0 +1,316 @@
...
+  const toc = document.getElementById('toc');
+
+  // eslint-disable-next-line prettier/prettier
+  specContainer.insertBefore(document.getElementById('sec-bibliography'), document.getElementById('sec-colophon'));
+  Array.from(toc.getElementsByTagName('a')).forEach(anchor => {
+    if (anchor.getAttribute('href') === '#sec-colophon') {
+      toc.getElementsByTagName('ol')[0].appendChild(anchor.parentNode);
+    }
+  });
+}
```
michaelficarra, line 285: I think it'd be fine to just rearrange these in the 262 spec document directly.

js/print.js
gesa: Which part? This function exists to gather information on sections that the links themselves don’t have (and CSS wouldn’t have access to, I don’t think?)

js/print.js
gesa: The document has the original copyright but not the "alternative copyright”. Would extremely love to be proven wrong on this though, it’s quite possible I’m missing something with copyright minutae.

css/elements.css
gesa: `metadata-block` does appear [in 402](https://github.com/tc39/ecma402/blob/ef423a80538a83e395e63068210778562de1209a/spec/metablock.html#L2). I thought unicode-property-table did too, but I was mistaken. I’ll make that edit.

js/print.js
gesa: of course it can. getting the scripts to output what i need is a finicky thing though and I try to feed it the simplest, cleanest markup i can.

js/print.js
gesa: I’d prefer to either leave that in your capable hands, or tackle this myself in a separate PR. Most importantly, after all the time, effort, and ecma money I dumped into producing a quality document, I want to merge what worked.

js/print.js
gesa: I’d like very much to address that in a separate PR as well.

src/H1.ts
```diff
@@ -20,11 +20,19 @@ export default class H1 extends Builder {
...
+
+      // const frag = spec.doc.createElement('div');
+      // frag.innerHTML = parent.getSecnumHTML() + node.innerHTML;
+      // node.prepend(...frag.children);
+      // node.innerHTML = ;
+      // const numElem = spec.doc.createElement('span');
+      // numElem.setAttribute('class', 'secnum');
+      // numElem.textContent = parent.number;
+      // node.insertBefore(spec.doc.createTextNode(' '), node.firstChild);
+      // node.insertBefore(numElem, node.firstChild);
```
michaelficarra, line 35: dead comments

css/print.css
```diff
@@ -84,119 +528,123 @@ emu-figure img {
...
 }
 
-emu-grammar + emu-alg,
-figcaption + emu-table,
-figcaption + span[id] + emu-table,
-emu-alg > ol > li:last-child {
-  break-before: avoid;
+.full-page-svg {
+  color: rgba(255, 255, 255, 0);
 }
```
michaelficarra, line 566: @gesa Kevin and I couldn't figure out what this class is used for. Why have transparent white text on the front cover and inner cover? And what's with the class name?

bakkot (approved):
Michael and to a lesser extent I messed with this a bunch, mostly because Michael has strong opinions about how this project should work, but we think it's behaviorally equivalent.

michaelficarra (approved):

PR #601, opened 2024-06-17
remove unnecessary break in SDOs reference pane header

michaelficarra:
before:

image

after:

image

### DIFF
--- a/js/menu.js
+++ b/js/menu.js
@@ -843,7 +843,7 @@ let referencePane = {
       e.parentNode.replaceChild(document.createTextNode(e.textContent), e);
     });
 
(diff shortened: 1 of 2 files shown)


bakkot:
I think this break is necessary for longer productions:
Screenshot 2024-06-17 at 7 37 45 AM

michaelficarra:
Breaking doesn't prevent it from wrapping on a narrow-enough display. All it does is wrap unnecessarily on larger displays. We may be able to add some CSS to prefer breaking before the grammar alternative.

michaelficarra:
Alternatively, we could just cut it off with an ellipsis when it doesn't fit:

image

bakkot:
If you want to do that and _also_ have the linebreak, I'm ok with it I guess. But I don't want to make things less readable than they currently are. The screenshot I posted looks fine; dropping the linebreak and adding an ellipsis would be worse.

PR #602, opened 2024-08-16
Suppress line breaks before field/slot references

gibson042:
I found an instance where a line break after the period in _specAlias_.[[SlotName]] made for confusing reading, making it look like a sentence was ending prematurely. This CSS change suppresses such line breaks by generating a non-copyable [word joiner](https://en.wikipedia.org/wiki/Word_joiner) before each field/slot reference.

<details><summary><b>Before</b></summary>

Note the line break after "[[Block]] is _buffer_." and before the "[[ArrayBufferByteLengthData]]" that is connected to _buffer_:

screenshot (before change)

</details>

<details><summary><b>After</b></summary>

Note the line break moving to after "[[Block]] is" and before "_buffer_.[[ArrayBufferByteLengthData]]", keeping the latter together as a single unit:

screenshot (after change)

</details>

### DIFF
--- a/css/elements.css
+++ b/css/elements.css
@@ -222,6 +222,10 @@ var.field {
   font: inherit;
   color: inherit;
 }
+/* suppress line break opportunities between `.` and `[[FieldName]]` */
+var.field::before {
+  content: '\2060'
+}
 
 var.referenced {
   color: inherit;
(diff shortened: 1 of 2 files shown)


bakkot (approved):
Thanks!

PR #603, opened 2024-08-16
Simplify buildMultipage

gibson042:


src/Spec.ts
```diff
@@ -740,36 +740,46 @@ export default class Spec {
...
         ruleId: 'top-level-section-id',
         message: 'When using --multipage, top-level sections must have ids beginning with `sec-`',
         node: ele,
       });
-      return false;
+      return undefined;
     }
-    if (!/^[A-Za-z0-9-_]+$/.test(ele.id)) {
+    const name = ele.id.substring(4);
+    if (!/^[A-Za-z0-9-_]+$/.test(name)) {
```
gibson042, line 763: This is technically a behavioral change, but we'd never consider `id="sec-"` to be valid anyway.

bakkot (approved):

PR #604, opened 2024-08-26
Add indentation to hidden counter text

gibson042:
This results in better copied text, e.g.
```
b. Set length to length + 1.
c. Let weight be GetWeight of node.
d. If length = 10**9 or weight > 2**(length + 1), then
   i. NOTE: This is an out-of-bounds state.
   ii. If ignoreErrors is not true, then
       1. Throw a RangeError exception.
   iii. Set hadError to true.
```
rather than
```
b. Set length to length + 1.
c. Let weight be GetWeight of node.
d. If length = 10**9 or weight > 2**(length + 1), then
i. NOTE: This is an out-of-bounds state.
ii. If ignoreErrors is not true, then
1. Throw a RangeError exception.
iii. Set hadError to true.
```

### DIFF
--- a/css/elements.css
+++ b/css/elements.css
@@ -125,6 +125,7 @@ body.oldtoc {
 
 span[aria-hidden='true'] {
   font-size: 0;
+  white-space: pre;
 }
 
 a {
--- a/js/listNumbers.js
+++ b/js/listNumbers.js
@@ -129,19 +129,30 @@ document.addEventListener('DOMContentLoaded', () => {
 
 // Omit indendation when copying a single algorithm step.
 document.addEventListener('copy', evt => {
-  const selection = getSelection();
-  const singleRange = selection?.rangeCount === 1 && selection.getRangeAt(0);
-  const container = singleRange?.commonAncestorContainer;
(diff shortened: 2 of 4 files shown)


bakkot:
Thanks! I don't like that the whitespace is included even if just copying a single line, though, as one does fairly often when writing an implementation which closely adheres to the spec text. I'm a bit hesitant to mess with the `copy` event, but it's probably appropriate here.

gibson042:
OK, that makes sense. Done.

bakkot:
I think that logic is broken in Chrome and Safari, which give slightly different selection ranges when triple-clicking.

gibson042:
@bakkot Updated; please retest and let me know about any issues you encounter.

bakkot:
Looks like it works for steps with substeps, but not steps without. Specifically, triple-clicking on the "This is an out-of-bounds state" line in the docs ends up with the following tree structure in `domRoot`, which doesn't trigger the conditions you're checking for:

Screenshot 2024-08-27 at 12 10 26 PM

gibson042:
Thanks for the detail. Fixed.

bakkot (approved):
Thanks!

Issue #606, opened 2024-09-12
emu-format --check only checks the first file on the command line


ptomato:
The help for `--check` says: "Exit with 1 if running with `--write` would cause at least some file to change (and print a list of such files); otherwise exit with 0."

The actual behaviour is that if the first file given on the command line needs formatting, its name is printed and the exit code is 1. If the first file doesn't need formatting, the remaining files are not examined and the exit code is 0.

PR #607, opened 2024-09-12
Fix emu-format --check with multiple files

ptomato:
Previously, running emu-format --check with multiple files would only check the first file. Fix the CLI arguments processing logic so that the only special case for a list of one file is when it needs to be printed to stdout. Move checking into the loop through the file list, as is done with writing.

Closes: #606


bakkot (approved):
Thanks!

Issue #608, opened 2024-09-12
Unexpected interaction between emu-format changing spacing around <ins> and completion records


ptomato:
When applying emu-format to Temporal there was one instance where the text produced by emu-format didn't pass ecmarkup --lint.

```diff
diff --git a/spec/intl.html b/spec/intl.html
index d0005fed..02578d4d 100644
--- a/spec/intl.html
+++ b/spec/intl.html
@@ -1017,7 +1017,7 @@
           1. Let _pattern_ be _format_.[[pattern12]].
         1. Else,
           1. Let _pattern_ be _format_.[[pattern]].
-        1. Let _result_ be <del>? </del><ins>! </ins>FormatDateTimePattern(_dateTimeFormat_, _format_, _pattern_, <del>_x_</del><ins>_xFormatRecord_.[[EpochNanoseconds]]</ins>).
+        1. Let _result_ be <del>?</del> <ins>!</ins> FormatDateTimePattern(_dateTimeFormat_, _format_, _pattern_, <del>_x_</del><ins>_xFormatRecord_.[[EpochNanoseconds]]</ins>).
         1. Return _result_.
       </emu-alg>
     </emu-clause>
```

Causes this error:

> Error: spec/intl.html:1020:54: FormatDateTimePattern returns a Completion Record, but is not consumed as if it does

I'm not sure if the fix should be to format `<ins>` and `<del>` differently, or to recognize this formatting as unwrapping the completion record.

This isn't a high priority problem for me, as I've found another way to work around it.

michaelficarra:
> I'm not sure if the fix should be to format `<ins>` and `<del>` differently, or to recognize this formatting as unwrapping the completion record.

I'm fine with fixing ecmarkup to handle this, but thinking as a reviewer, I'd probably rather just see full steps added/removed.

PR #610, opened 2024-09-12
handle ins/del in completion-record typechecks

bakkot:
Fixes https://github.com/tc39/ecmarkup/issues/608.

The typechecker is already using the expr-parser output, which strips ins/del, but the result of that parse can have multiple adjacent `name: 'text'` items in sequence, which the typechecker didn't account for here. Specifically, with the input `<ins>?</ins> Foo()` we end up with `[{ name: 'text', content: '?' }, { name: 'text', content: ' ' }, foo]`. Adjust the logic to handle this case.


michaelficarra:
Let's add a test for using both `ins` and `del` to change the sigil, like in the original issue.

michaelficarra (approved):

PR #611, opened 2024-09-17
Bump engines and some dependencies

bakkot:
This is a breaking change, but node 16 is no longer supported, so whatever. We claimed to support node 8 but haven't been testing on it in forever.

This isn't quite the latest typescript because typescript-eslint hasn't updated to support it yet.


package.json
```diff
@@ -72,22 +72,22 @@
...
+		"typescript": "^5.5.4"
 	},
 	"prettier": {
 		"singleQuote": true,
 		"arrowParens": "avoid",
 		"printWidth": 100
 	},
 	"engines": {
-		"node": ">= 12 || ^11.10.1 || ^10.13 || ^8.10"
+		"node": ">= 18"
```
ljharb, line 91: even if nothing else changed, this alone requires a semver-major bump, fwiw

bakkot: Yeah, that's why I said "This is a breaking change".

tsconfig.test.json
```diff
@@ -2,6 +2,7 @@
     "extends": "./tsconfig",
     "include": ["lib/**/*"],
     "compilerOptions": {
+        "outDir": "this needs to be set so it does not ignore lib",
```
michaelficarra, line 5: wow

michaelficarra (approved):

package.json
ljharb: gotcha, the "so whatever" implied you weren't going to publish it as one.

bakkot: Aha, I see how that was misleading; I just meant that it's unlikely to break anyone in practice.

PR #612, opened 2024-09-17
Refactor typechecker

bakkot:
Based on https://github.com/tc39/ecmarkup/pull/611.

No functional changes except for the error messages in some cases. Commits can be reviewed individually, if you really want to, but it's not interesting other than the "[split completion types in typechecker](https://github.com/tc39/ecmarkup/commit/51095be)" commit, which splits the "completion" type into normal and abrupt. They can be union'd like any other types to represent "either a normal completion or an abrupt completion", instead of needing to keep track of it as a kind of ad-hoc union.

We don't refine "abrupt" because there's no point but it would be easy to do from here if it becomes necessary.


src/type-logic.ts
```diff
@@ -0,0 +1,576 @@
...
+  return { kind: 'unknown' };
+}
+
+export function isCompletion(
+  type: Type,
+): type is Type & { kind: 'normal completion' | 'abrupt completion' | 'union' } {
+  return (
+    type.kind === 'normal completion' ||
+    type.kind === 'abrupt completion' ||
+    (type.kind === 'union' && type.of.some(isCompletion))
```
michaelficarra, line 559: You can't mix completions and non-completions, right? So shouldn't this just check `isCompletion(type.of[0])`?

michaelficarra (approved):
LGTM otherwise

src/type-logic.ts
bakkot: Not syntactically, but it's possible for it to arise as a result of (more sophisticated than currently implemented) type analysis. For example, consider

```
1. Let _x_ be Completion(_something_).
1. If _x_ is a normal completion], then
  1. Set _x_ to ! _x_.
1. Return _x_.
```

On the last line, if we tracked the types of variables, we might think `_x_` is either a completion or not. The user shouldn't write this program but the analysis shouldn't go off the rails if they do.

michaelficarra: Ah right, we arrive at this through type analysis, not just type declarations. 👍

PR #613, opened 2024-09-17
special handling for Completion/NormalCompletion in the typechecker

bakkot:
Based on https://github.com/tc39/ecmarkup/pull/612.

We don't (and are not going to) have generic types, so just hardcode these specific generic types. Also observe that a normal completion of `never` is equivalent to `never`, which allows us to get errors when passing `NormalCompletion(x)` to `NormalCompletion(not-x)`. The latter doesn't really come up for arguments but it does for returned values, which I want to handle soon.

### DIFF
--- a/src/type-logic.ts
+++ b/src/type-logic.ts
@@ -322,7 +322,11 @@ export function serialize(type: Type): string {
   }
 }
 
-export function typeFromExpr(expr: Expr, biblio: Biblio): Type {
+export function typeFromExpr(
+  expr: Expr,
+  biblio: Biblio,
+  warn: (offset: number, message: string) => void,
+): Type {
   seq: if (expr.name === 'seq') {
     const items = stripWhitespace(expr.items);
     if (items.length === 1) {
@@ -360,7 +364,7 @@ export function typeFromExpr(expr: Expr, biblio: Biblio): Type {
     if (items[0]?.name === 'text' && ['!', '?'].includes(items[0].contents.trim())) {
       const remaining = stripWhitespace(items.slice(1));
       if (remaining.length === 1 && ['call', 'sdo-call'].includes(remaining[0].name)) {
-        const callType = typeFromExpr(remaining[0], biblio);
+        const callType = typeFromExpr(remaining[0], biblio, warn);
         if (isCompletion(callType)) {
           const normal: Type =
             callType.kind === 'normal completion'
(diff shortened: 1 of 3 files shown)


src/type-logic.ts
```diff
@@ -398,6 +402,29 @@ export function typeFromExpr(expr: Expr, biblio: Biblio): Type {
...
+              { kind: 'abrupt completion' },
+            ],
+          };
+        }
+      }
+      // special case: `NormalCompletion` wraps its input
+      if (
+        expr.name === 'call' &&
+        calleeName === 'NormalCompletion' &&
+        expr.arguments.length === 1
```
michaelficarra, line 423: I think these arguments length checks should be assertions, not tests.

src/type-logic.ts
```diff
@@ -207,11 +207,15 @@ export function meet(a: Type, b: Type): Type {
...
-  ) {
-    return { kind: a.kind, of: meet(a.of, b.of) };
+  if (a.kind === 'list' && b.kind === 'list') {
+    return { kind: 'list', of: meet(a.of, b.of) };
+  }
+  if (a.kind === 'normal completion' && b.kind === 'normal completion') {
+    const inner = meet(a.of, b.of);
+    if (inner.kind === 'never') {
+      return { kind: 'never' };
+    }
```
michaelficarra, line 217: Doesn't this same thing apply to `list` above? Why handle these differently?

michaelficarra (commented):
We should add an `assertNoTypeError` test for `a normal completion containing an ECMAScript language value` and `NormalCompletion(*true*)` or something (the normal completion parameter type being a subtype of the declared type).

src/type-logic.ts
bakkot: The empty list is an inhabitant of `List<never>`, but you can't have an empty normal completion, so there are no inhabitants of `NormalCompletion<never>`.

Before we split "abrupt completion" and "normal completion" into distinct types (in the previous PR) there was just `Completion<never>`, which held abrupt completions, but now we can do better. Probably this change should have been in the previous PR but whatever.

michaelficarra: Ah I forgot about the empty list! 😖

src/type-logic.ts
bakkot: Done. I don't think it was worth my time to make that change, but too late now.

bakkot:
> We should add an `assertNoTypeError` test for `a normal completion containing an ECMAScript language value` and `NormalCompletion(*true*)` or something (the normal completion parameter type being a subtype of the declared type).

Done. Though our type system represents `false` as a distinct type from `Boolean`, so strictly speaking [we already had that](https://github.com/tc39/ecmarkup/blob/540ad3a08f269a133f4383c1aa273dec97e80b83/test/typecheck.js#L1542-L1546).

michaelficarra:
I guess we could simplify that other test to `a normal completion containing *false*` if we wanted.

michaelficarra (approved):

Issue #609, opened 2024-09-12
Deleted lines should not be considered to return a completion record from an AO


ptomato:
Test case:
```html
<emu-clause id="sec-foobar" type="abstract operation">
  <h1>
    FooBar (
      _v_: an ECMAScript language value,
    ): <ins>an integer</ins><del>either a normal completion containing an integer, or a throw completion</del>
  </h1>
  <dl class="header">
    <dt>description</dt>
    <dd></dd>
  </dl>
  <emu-alg>
    1. <del>Throw a *RangeError* exception.</del>
    1. Return 42.
  </emu-alg>
</emu-clause>
```

This produces:

> (file):5:23: unknown extra text in header
> (file):12:13: this would return a Completion Record, but the containing AO is declared not to return a Completion Record

Looks like there is something that needs to be handled differently about the structured header return type, as well.

PR #615, opened 2024-09-17
Check return types

bakkot:
Based on https://github.com/tc39/ecmarkup/pull/613.

This will give errors for cases like

```js
<emu-clause id="example" type="abstract operation">
  <h1>
    Example (): a Number
  </h1>
  <dl class="header"></dl>
  <emu-alg>
    1. Return 0.
  </emu-alg>
</emu-clause>
```
such as
```
Error: test.emu:13:23: returned value (0) does not look plausibly assignable to algorithm's return type (Number)
hint: you returned a mathematical value, but this algorithm returns an ES language Number
```

Fixes https://github.com/tc39/ecmarkup/issues/609 by moving the logic for checking for returned abrupt completions out of an ad-hoc bit of `Algorithm.ts` and into the typechecker as part of a more general facility for checking the return types from algorithms.

This also drops the special handling for `GeneratorYield` and instead adds a new attribute, "skip return checks", which will need to be added to it as well as `Completion` and `NormalCompletion` and any other algorithms which are doing weird stuff in the future. As such, this is a **breaking change** (but only very slightly).

This introduces no new errors on ecma262, ecma402, or proposal-temporal except for the three algorithms called out in the previous paragraph, which will be fixed by adding the new attribute when upstreaming this change.

The actual logic, in `inspectReturns`, is pleasingly simple: we just look for `Return Foo.`, parse `Foo` the same way we parse arguments to AO calls, and compare its type to the algorithm's return type the same way we'd compare a call argument to the corresponding AO's corresponding parameter type. We also look for `?` and `Throw` and etc so we can ensure the containing algorithm returns an abrupt completion.


src/type-logic.ts
```diff
@@ -571,7 +574,7 @@ export function typeFromExprType(type: BiblioType): Type {
     }
     case 'unused': {
       // this is really only a return type, but might as well handle it
```
michaelficarra, line 576: This comment no longer makes sense now that we handle return types.

```suggestion
```

spec/index.html
```diff
@@ -212,6 +212,7 @@ <h1>Structured Headers</h1>
             <li><b>for:</b> The type of value to which a clause of type "concrete method" or "internal method" applies.</li>
             <li><b>redefinition:</b> If "true", the name of the operation will not automatically link (i.e., it will not automatically be given an aoid).</li>
             <li><b>skip global checks:</b> If "true", disables consistency checks for this AO which require knowing every callsite.</li>
+            <li><b>skip return checks:</b> If "true", disables checking that the returned values from this AO correspond to its declared return type.</li>
```
michaelficarra, line 215: Instead of skipping return checks, shouldn't we be asserting that the return check fails? That way if it is later changed in such a way that it passes, we know we can remove the annotation.

src/typechecker.ts
```diff
@@ -179,6 +185,37 @@ const getExpressionVisitor =
     }
   };
 
+type ErrorForUsingTypeXAsTypeY =
+  | null
+  | 'number-to-math'
```
michaelficarra, line 190: We use `real` elsewhere, not `math`.

michaelficarra, line 190: For completion's sake, shouldn't we have `bigint-to-real`?

src/typechecker.ts
```diff
@@ -77,17 +81,31 @@ const getExpressionVisitor =
...
+              hint =
+                '\nhint: you passed a mathematical value, but this position takes an ES language Number';
+              break;
+            }
+            case 'math-to-bigint': {
+              hint =
+                '\nhint: you passed a mathematical value, but this position takes an ES language BigInt';
+              break;
+            }
+            case 'other': {
```
michaelficarra, line 103: Instead of testing for `null` above, either use a `default` or (if you're looking to get an exhaustivity check from TS), fallthrough with `case null:` here.

src/typechecker.ts
```diff
@@ -179,6 +185,37 @@ const getExpressionVisitor =
     }
   };
 
+type ErrorForUsingTypeXAsTypeY =
+  | null
```
michaelficarra, line 189: In what way does `null` represent something different from `'other'`? You should document that here in a comment.

src/typechecker.ts
```diff
@@ -332,6 +435,152 @@ function isConsumedAsCompletion(expr: Expr, path: PathItem[]) {
   return false;
 }
 
+// returns a boolean to indicate whether this line can return an abrupt completion, or null to indicate we should stop caring
```
michaelficarra, line 438: I don't really like that. Can't we make an enum for it?

michaelficarra (commented):
I think we're missing tests where the completiony-ness of an Abstract Closure doesn't match that of the containing algorithm (in both ways).

src/typechecker.ts
bakkot: Then we couldn't use `||=`.

michaelficarra: just implement `Functor` 😛

bakkot:
> I think we're missing tests where the completiony-ness of an Abstract Closure doesn't match that of the containing algorithm (in both ways).

We have an assertion that this is not an error:

```
<emu-clause id="example" type="abstract operation">
<h1>
  ExampleAlg (): a mathematical value
</h1>
<dl class="header">
</dl>
<emu-alg>
  1. Let _addend_ be 41.
  1. Let _closure_ be a new Abstract Closure with parameters (_x_) that captures _addend_ and performs the following steps when called:
    1. Return _x_ + ? ExampleCompletionAO(_addend_).
  1. Let _val_ be ! _closure_(1).
  1. Return _val_.
</emu-alg>
</emu-clause>
```

What else are you looking for?

src/typechecker.ts
bakkot: no

michaelficarra:
Ah that's right, I misread this test and thought they were both returning completion records. Then the only one we're missing is one where the algorithm returns a completion but the AC returns a non-completion.

bakkot:
> Then the only one we're missing is one where the algorithm returns a completion but the AC returns a non-completion.

That wouldn't be an error even with no handling of ACs because values returned from AOs which return completions are automatically boxed into NormalCompletion.

michaelficarra:
I wouldn't expect the Return steps inside an AC to auto-wrap in NormalCompletion just because the containing algorithm is declared to return a completion. If that's how it currently works, it shouldn't be.

spec/index.html
bakkot: That would probably be better but I don't want to do the work to implement it. PRs welcome.

src/typechecker.ts
bakkot: Sure I guess. That's not related to this PR but I can fold it in.

src/typechecker.ts
bakkot: The test for `null` guards the whole error-producing logic, not just this switch case.

bakkot:
> I wouldn't expect the Return steps inside an AC to auto-wrap in NormalCompletion just because the containing algorithm is declared to return a completion. If that's how it currently works, it shouldn't be.

It's not, but ACs don't have return types. The bodies of ACs are exempt from the return type analysis entirely.

My point is that if the body incorrectly _weren't_ exempt from the analysis, such that a `return` in the body of the AC were to be incorrectly considered to be returning from the containing AO, and we had e.g. `return 0` in an AO declared to return a normal completion containing a real number, that _still_ wouldn't be an error (because of the auto-boxing logic).

So there's no test you could write here which would fail under any code we could have written. So there's no reason to have a test.

src/typechecker.ts
bakkot: Done. (It's "no error".)

src/type-logic.ts
bakkot: Done.

src/typechecker.ts
bakkot: Done.

spec/index.html
bakkot: OK actually that was trivial, done.

michaelficarra (approved):
I might call this something else now, but LGTM either way.

jmdyck:
> I wouldn't expect the Return steps inside an AC to auto-wrap in NormalCompletion just because the containing algorithm is declared to return a completion. If that's how it currently works, it shouldn't be.

The spec isn't entirely clear on this point. 5.2.3.5 "Implicit Normal Completion" says "In algorithms within abstract operations which are declared to return a Completion Record, ..." which would seem to include an AC within such an AO.

Maybe that section should carve out an exception for ACs. But then the question might arise as to whether ACs have their own rule for implicit `NormalCompletion`. Some seem to be written under the assumption of auto-wrapping.

(Tangent: maybe ACs should explicitly state their return type.)

PR #626, opened 2024-11-21
add a meta tag to set viewport options for mobile devices

michaelficarra:
Originally from https://github.com/tc39/ecma262/pull/3481.


brettearle:
Nice work, it was good seeing where you did this in here. I looked and thought such a big sweeping change to baselines might be a problem.

test/baselines/generated-reference/algorithm-replacements.html
```diff
@@ -1,5 +1,5 @@
 <!doctype html>
-<head><meta charset="utf-8"></head><body><div id="shortcuts-help">
+<head><meta name="viewport" content="width=device-width, initial-scale=1"><meta charset="utf-8"></head><body><div id="shortcuts-help">
```
bakkot, line 2: Kind of unfortunate this ends up being the second element but it's still close enough to the beginning of the file so it's probably fine.

bakkot (approved):

PR #637, opened 2025-03-10
Fix 'oneof' in syntax-directed operations

szuend:
In the source map spec we have the use-case of declaring an SDO over a lexical production (for VLQ decoding).

This PR adds the missing implementation.

### DIFF
--- a/src/lint/utils.ts
+++ b/src/lint/utils.ts
@@ -75,6 +75,9 @@ export function rhsMatches(a: RightHandSide | OneOfList, b: RightHandSide | OneO
     }
     case SyntaxKind.OneOfList:
       return oneOfListMatches(a, b as OneOfList);
+    default:
+      // @ts-expect-error: Callers might pass other kinds of nodes.
+      throw new Error('unknown rhs type ' + a.constructor.name);
   }
 }
 
--- a/src/lint/utils.ts
+++ b/src/lint/utils.ts
@@ -175,10 +175,8 @@ function oneOfListMatches(a: OneOfList, b: OneOfList) {
   if (a.terminals === undefined || b.terminals === undefined) {
     throw new Error('OneOfList must have terminals');
   }
-  return (
-    a.terminals.length === b.terminals.length &&
-    a.terminals.every((ae, i) => ae.text === b.terminals![i].text)
-  );
+  // The terminals in a must be a subset of the terminals in b.
+  return a.terminals.every(ae => b.terminals!.find(be => ae.text === be.text));
 }
 
 // this is only for use with single-file grammars
--- a/src/lint/utils.ts
+++ b/src/lint/utils.ts
@@ -73,8 +73,8 @@ export function rhsMatches(a: RightHandSide | OneOfList, b: RightHandSide | OneO
       }
       return symbolSpanMatches(aHead, bHead);
     }
-    default:
-      throw new Error('unknown rhs type ' + a.constructor.name);
+    case SyntaxKind.OneOfList:
+      return oneOfListMatches(a, b as OneOfList);
   }
 }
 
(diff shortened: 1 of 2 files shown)


test/errors.js
```diff
@@ -1106,6 +1106,31 @@ ${M}      </pre>
...
+            Foo :: ${M}one of \`a\`
+          </emu-grammar>
+          <emu-alg>
+            1. Return *true*.
+          </emu-alg>
+        </emu-clause>`,
+        {
+          ruleId: 'grammar-shape',
+          nodeType: 'emu-grammar',
+          message: 'could not find definition for rhs "a"',
```
nicolo-ribaudo, line 1129: Why is this reporting an error, if `a` is one of the valid values?

szuend: Apologies, it's late here. I guess what we actually want is a subset check?

nicolo-ribaudo: Probably, yes 👍

szuend: Done, also added a negative test to cover this.

test/errors.js
```diff
@@ -1167,5 +1192,49 @@ ${M}      </pre>
...
+      await assertErrorFree(`
+        <emu-grammar type="definition">
+          Foo :: one of \`a\` \`b\` \`c\`
+        </emu-grammar>
+
+        <emu-clause id="sec-example" type="sdo">
+        <h1>Static Semantics: Example</h1>
+        <dl class='header'></dl>
+          <emu-grammar>
+            Foo :: one of \`a\`
```
michaelficarra, line 1225: This should read

```suggestion
            Foo :: \`a\`
```

We should prohibit "one of" with only one alternative.

nicolo-ribaudo: Do you mean that
```
<emu-gramar>
  Foo ::
    `a`
    `b`
    `c`
    `d`
</emu-grammar>
<emu-clause>
  1. Return 1.
</emu-clause>
```
is preferable to
```
<emu-gramar>
  Foo :: one of `a` `b` `c` `d`
</emu-grammar>
<emu-clause>
  1. Return 1.
</emu-clause>
```
?

nicolo-ribaudo: For completeness sake, the case that triggered this PR would be
```
      <emu-grammar>
        TerminalDigit :: one of
          `A` `B` `C` `D` `E` `F` `G` `H` `I` `J` `K` `L` `M` `N` `O` `P` `Q` `R`
          `S` `T` `U` `V` `W` `X` `Y` `Z` `a` `b` `c` `d` `e` `f`
      </emu-grammar>
      <emu-alg>
        1. Let _digit_ be the source text matched by this Parse Node.
        1. ...
```

jmdyck:
Note that the current spec does have examples of an SDO being defined on a "one of" production, but they're all in the 'inline' form, rather than the `emu-grammar`+`emu-alg` form:
```
- The TRV of <emu-grammar>DecimalDigit :: one of `0` `1` `2` `3` `4` `5` `6` `7` `8` `9`</emu-grammar> is ...
- The TRV of <emu-grammar>HexDigit :: one of `0` `1` `2` `3` `4` `5` `6` `7` `8` `9` `a` `b` `c` `d` `e` `f` `A` `B` `C` `D` `E` `F`</emu-grammar> is ...
- The TRV of <emu-grammar>SingleEscapeCharacter :: one of `'` `"` `\` `b` `f` `n` `r` `t` `v`</emu-grammar> is ...
```

test/errors.js
michaelficarra: 🤔 okay I'm convinced.

szuend: I changed the test to ``Foo :: one of `a` `b` `` instead. Allowing ``Foo :: `a` `` should probably be done in a separate PR since it requires changing `rhsMatches` to compare different kinds of productions. Also disallowing ``Foo :: one of `a` `` should probably be done in the linter in a separate PR.

src/lint/utils.ts
```diff
@@ -73,8 +73,8 @@ export function rhsMatches(a: RightHandSide | OneOfList, b: RightHandSide | OneO
       }
       return symbolSpanMatches(aHead, bHead);
     }
-    default:
-      throw new Error('unknown rhs type ' + a.constructor.name);
+    case SyntaxKind.OneOfList:
+      return oneOfListMatches(a, b as OneOfList);
```
bakkot, line 77: This needs a `default`. (If TS complains, add a `ts-expect-error`.)

I can do this myself when landing if you don't get to it first.

szuend: Done

szuend:
Anything missing to get this landed?

bakkot (approved):

bakkot:
Published in 21.1.0, sorry for the delay.

PR #462, opened 2022-06-22
Implement alternative license text

codehag:

PR #638, opened 2025-03-12
Some fixes to maintain compliance to the styleguide

gesa:
More TCs and TGs are expressing an interest in Ecmarkup, so I want to make sure we can produce the mandatory documents for them at standardization time.

Man I really hate the logic I came up with for the copyright page. Alternative: make sure it gets an easy-to-reference ID and move that back into the print.js DOM manipulation.

~~This is a draft because I’ve only glossed over the multipage output, haven’t given it a good look yet.~~

This has all been checked on printable & multi-page ECMA-262 & ECMA-402 builds. I noticed no changes on multipage builds.

## fixed

- no longer failing on draft status due to code removing shortname after checking for its presence

<details>
<summary>replaced blank cover</summary>

2503_s9qwa

</details>

<details>
<summary>removed unexpected title page</summary>

2503_luqn0

</details>

- removed extraneous address addition to inside cover (and its associated style)
- added missing license page from ToC (it should be listed the same way the bibliography or colophon are)

<details><summary>replaced incorrectly formatted copyright notice—should match template</summary>

2503_c2xwo

</details>

- software license in wrong spot for printed standards moved to end
- added correct copyright notice for certain standards ([Ecma “alternative” copyright notice](https://ecma-international.org/policies/by-ipr/ecma-text-copyright-policy/#alternative-copyright-notice))
  - `alternative` option added to `boilerplate.copyright` for documents which use Ecma’s alternative copyright

<details><summary>formatted Annex titles as expected by template</summary>

2503_iu5qm

</details>

## needs extra scrutiny

I removed the code that moves the “scope” section to the beginning when printable option is passed. I didn’t see a spec that had the scope anywhere else, nor would I expect it. Maybe left over from when I was doing more manipulation on the prince side? Need a gut check that it’s good to go.

<details>

<summary>I changed the special treatment that the shortname gets when there’s a draft, as it breaks anything relying on the shortname being present. It shouldn’t impact multipage builds at all, but I suppose I could imagine a corner case where before changes to after</summary>

Before:

2503_3t3f1

After: 

2503_v0nly

</details>

The generated references are meaningfully different because of 👆🏻.  Seeking an opinion on whether or not they should not be.

### DIFF
--- a/PUBLISHING.md
+++ b/PUBLISHING.md
@@ -3,21 +3,42 @@
 ## Required frontmatter
 
 - In order to produce a PDF, the front matter `title`, `shortname`, and `status` are **mandatory**.
-- If using the alternative copyright, `location` is also mandatory.
-- If generating a version for submission to the GA, `version` and `date` are mandatory. `date` should reflect the date of the Ecma GA which will ratify the Standard. For example:
 
+```yaml
+title: Temporal proposal
+shortname: Temporal
+status: proposal
+stage: 3
 ```
-title: ECMAScript® 2024 Language Specification
+
+- If using the alternative copyright for a standard publication, `location` is also mandatory. For example:
+
+```yaml
+title: ECMAScript® Language Specification
 shortname: ECMA-262
-version: 15<sup>th</sup> Edition
-date: 2024-06-25
+status: draft
+boilerplate:
+  copyright: alternative
+```
+
+- If generating a version for submission to the GA, `version` and `date` are mandatory. `date` should reflect the date of the Ecma GA which will ratify the Standard.
+
+```yaml
+title: ECMAScript® 2025 Language Specification
+shortname: ECMA-262
+version: 16<sup>th</sup> Edition
+date: 2025-06-25
+status: standard
+boilerplate:
+  copyright: alternative
+location: https://262.ecma-international.org/16.0/
 ```
 
 To generate markup for use in PDF conversion, make sure to include the options `--assets`, `--assets-dir`, and `--printable`. If you have images and styles to include, make sure to move them into your assets directory before running `ecmarkup`. For example:
 
 ```shell
-mkdir -p out &&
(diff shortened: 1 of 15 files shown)


michaelficarra:
Can you explain a bit more about the situation? What is required and what are we doing now that fails to meet the requirement? Also, if it would help, provide screenshots.

michaelficarra:
It looks like the default copyright in ecmarkup is an earlier version of the current default copyright listed at https://ecma-international.org/policies/by-ipr/ecma-text-copyright-policy/. We should also update that while we're at it.

gesa:
good call

michaelficarra:
From reading https://github.com/tc39/ecmarkup/pull/462, it seems like we want all TC39 documents (specs and draft specs) to have only the alternative license. And I'm assuming other TCs will use the default copyright. Is that correct?

gesa:
@michaelficarra I’m waiting to hear back from the secretariat which (if any) other TCs have requested to use the alternative copyright from the GA

PUBLISHING.md
```diff
@@ -1,6 +1,10 @@
...
 
-In order to produce a PDF, the front matter `title`, `shortname`, `version`, and `date` are **mandatory**. If generating a final annual edition, date should reflect the date of the Ecma GA which will ratify the Standard. For example:
+## Required frontmatter
+
+- In order to produce a PDF, the front matter `title`, `shortname`, and `status` are **mandatory**.
+- If using the alternative copyright, `location` is also mandatory.
+- If generating a version for submission to the GA, `version` and `date` are mandatory. `date` should reflect the date of the Ecma GA which will ratify the Standard. For example:
 
 ```
 title: ECMAScript® 2024 Language Specification
```
michaelficarra, line 10: This example should be updated to include status and location.

bakkot (approved):

PR #641, opened 2025-04-18
Fix display of normative optional blocks used as algorithm steps

### LINKED ISSUE #640: Normative optional blocks used as algorithm steps no longer display correctly
Looks like the addition of `display: block;` to `[normative-optional]` in #600 regressed display of normative optional blocks used as algorithm steps.

Currently,
```html
<emu-alg>
  1. [normative-optional] If the host is a web browser or otherwise supports <emu-xref href="#sec-late-errors-for-function-call-assignment-targets" title></emu-xref>, then
    1. If IsStrict(this |CallExpression|) is *false*, return ~web-compat~.
  1. Return ~invalid~.
</emu-alg>
```
produces
Image
wherein the `<li>` holding the normative optional block doesn't (visually) get assigned a number.

But in the [earlier-generated preview for PR 2952](https://ci.tc39.es/preview/tc39/ecma262/pull/2952#sec-toboolean), we can see the expected behavior:
Image

rkirsling:
Fixes #640.

Or at least, I hope it does—this reverts one line of #600, but I'm not sure what situation that line was meant to address. At a glance, this seems to fix the problem without having any effect on normative optional blocks which _aren't_ used as algorithm steps.

### DIFF
--- a/css/elements.css
+++ b/css/elements.css
@@ -1572,7 +1572,6 @@ emu-normative-optional,
 [deprecated],
 [legacy] {
   border-left: 5px solid var(--normative-optional-border-color);
-  display: block;
   padding: 0.5em;
   background: var(--normative-optional-background-color);
 }
--- a/css/elements.css
+++ b/css/elements.css
@@ -1561,6 +1561,10 @@ li.menu-search-result-term::before {
   padding-right: 5px;
 }
 
+emu-normative-optional {
+  display: block;
+}
+
 emu-normative-optional::before {
   display: block;
   color: var(--attributes-tag-foreground-color);
--- a/test/baselines/generated-reference/assets-inline.html
+++ b/test/baselines/generated-reference/assets-inline.html
@@ -3184,7 +3184,6 @@ emu-normative-optional,
 [deprecated],
 [legacy] {
   border-left: 5px solid var(--normative-optional-border-color);
-  display: block;
   padding: 0.5em;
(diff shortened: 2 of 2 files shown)


michaelficarra:
LGTM but let's see if @gesa knows why this line was added first.

rkirsling:
Okay, I see what happened:
- ECMA-262 and the ecmarkup docs do not use the `<emu-normative-optional>` element; the `normative-optional` attribute is consistently used instead.
- ECMA-402, on the other hand, exclusively uses `<emu-normative-optional>`, and the styles in question appear to have come from ECMA-402's [index.html style block](https://github.com/tc39/ecma402/blob/main/spec/index.html#L12-L22).

So I'll update this PR such that `display: block;` only applies to `<emu-normative-optional>`.

(Still, I wonder if ECMA-402 should be migrated from the element to the attribute?)

michaelficarra:
> * ECMA-402, on the other hand, exclusively uses `<emu-normative-optional>`,

I didn't even know that was a thing. Is that a thing? I don't see it in the ecmarkup docs. It certainly seems like 402 should migrate to the attribute as used in 262.

michaelficarra:
We talked about it in editor call today and we would indeed prefer 402 to replace its uses of an `<emu-normative-optional>` tag with the `normative-optional` attribute on the relevant ecmarkup element. I'll open an issue.

bakkot (approved):
Thanks!

PR #643, opened 2025-04-26
Balance padding of inline <emu-grammar>

rkirsling:
Currently, inline usage of `<emu-grammar>` has twice as much padding on the left as the right:
image
CSS-wise, we're using `padding-left: 1ex;` in this case, whereas the rest of the padding / margin involved in `<emu-grammar>` appears to use a value of `0.5ex` instead.

I guess it's been this way for years, but it really looks like a visual bug, so I'm hereby proposing we make the left side `0.5ex` as well, such that we get the following instead:
image


michaelficarra:
I actually kinda prefer the extra separation that 1ex gives us. Can you also provide a screenshot comparison of a balanced 1ex padding?

rkirsling:
Adding `padding-right: 0.5ex;` to the status quo `emu-production` (such that that adds onto the 0.5ex right margin for elements within it) looks like this:
image

...which I think looks fine in that particular case but is somewhat less great (IMO) when followed by a comma:
image

syg (approved):

michaelficarra (approved):

PR #645, opened 2025-04-29
add new keyboard shortcut to jump to the most recent link target

michaelficarra:
This adds `` ` `` as a new keyboard shortcut to jump to the last place you jumped. I like to think of it as a kind of "dynamic pin". When I click a link to somewhere in the spec, I often scroll or Ctrl-F my way away without thinking of pinning it and want a quick way to get back to what I was just looking at. I chose `` ` `` for the key because it nicely aligns with the numbered pins on my keyboard and the above-stated mental model as another pin. Also, it's the same key used to jump to a mark in vim.

### DIFF
--- a/js/menu.js
+++ b/js/menu.js
@@ -321,10 +321,14 @@ function Menu() {
 
 Menu.prototype.documentKeydown = function (e) {
   e.stopPropagation();
-  if (e.keyCode === 80) {
+  if (e.key === 'p') {
     this.togglePinEntry();
-  } else if (e.keyCode >= 48 && e.keyCode < 58) {
+  } else if ([0, 1, 2, 3, 4, 5, 6, 7, 8, 9].includes(parseInt(e.key))) {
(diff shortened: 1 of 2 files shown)


js/menu.js
```diff
@@ -321,10 +321,14 @@ function Menu() {
...
   e.stopPropagation();
-  if (e.keyCode === 80) {
+  if (e.key === 'p') {
     this.togglePinEntry();
-  } else if (e.keyCode >= 48 && e.keyCode < 58) {
+  } else if ([0, 1, 2, 3, 4, 5, 6, 7, 8, 9].includes(parseInt(e.key))) {
     this.selectPin((e.keyCode - 9) % 10);
+  } else if (e.key === '`') {
+    if (document.location.hash) {
+      document.querySelector(document.location.hash).scrollIntoView(true);
```
michaelficarra, line 330: This could also be done with `document.location = document.location`, but that felt gross to me.

bakkot (approved):

PR #646, opened 2025-04-29
Decode target ID from location.hash rather than using it as an ID selector

gibson042:
Ref #645

Code hygiene—there's no guarantee that `location.hash` is a valid CSS selector, or that it actually corresponds with an element.

### DIFF
--- a/test/baselines/generated-reference/assets-inline.html
+++ b/test/baselines/generated-reference/assets-inline.html
@@ -410,9 +410,10 @@ Menu.prototype.documentKeydown = function (e) {
   } else if ([0, 1, 2, 3, 4, 5, 6, 7, 8, 9].includes(parseInt(e.key))) {
     this.selectPin((e.keyCode - 9) % 10);
   } else if (e.key === '`') {
-    if (document.location.hash) {
-      document.querySelector(document.location.hash).scrollIntoView(true);
-    }
+    const hash = document.location.hash;
+    const id = decodeURIComponent(hash.slice(1));
+    const target = document.getElementById(id);
+    target?.scrollIntoView(true);
   }
 };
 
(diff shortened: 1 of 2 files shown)


ctcpip (approved):

test/baselines/generated-reference/assets-inline.html
```diff
@@ -410,9 +410,10 @@
   } else if ([0, 1, 2, 3, 4, 5, 6, 7, 8, 9].includes(parseInt(e.key))) {
     this.selectPin((e.keyCode - 9) % 10);
   } else if (e.key === '`') {
-    if (document.location.hash) {
-      document.querySelector(document.location.hash).scrollIntoView(true);
-    }
+    const hash = document.location.hash;
+    const id = decodeURIComponent(hash.slice(1));
+    const target = document.getElementById(id);
+    target?.scrollIntoView(true);
```
michaelficarra, line 416:
```suggestion
    if (id !== '') {
      const target = document.getElementById(id);
      target?.scrollIntoView(true);
    }
```

in case there's an element with the empty string for an ID I guess

michaelficarra (approved):
When did we start trying to make ecmarkup code good?

test/baselines/generated-reference/assets-inline.html
gibson042: There's no need, the DOM spec implements an invariant that the [unique identifier](https://dom.spec.whatwg.org/#concept-id) searched for by [`getElementById`](https://dom.spec.whatwg.org/#dom-nonelementparentnode-getelementbyid) is never empty (e.g., setting `id` to an empty string unsets the associated ID).

(_[demonstration](https://jsfiddle.net/54ef16b8/)_)

PR #651, opened 2025-09-06
fix https://github.com/tc39/ecma262/issues/3686

zhangenming:
e551da9c31760b6c3ee31b8bc392e4d8


ljharb:
While this change is fine, as I commented in https://github.com/tc39/ecma262/issues/3686, this seems like a flaw in whichever browser you're using.

zhangenming:
fine, I try chrome/edge/firefox , did not work.
but firefox work well

bakkot:
Can't figure out why this was inline-block in the first place but it doesn't seem to make a visual difference as far as I can tell.

PR #654, opened 2025-09-12
Fix: print styles for web visitors

gesa:
I got a little nerd sniped on Mastodon by a conversation about accessibility and print styles. Browsers mostly don't implement the CSS paged media module, and the web version of typical ecmarkup output doesn't use the oldtoc class, so I needed to add just a few little things to make the site printable for web visitors.

> [!IMPORTANT]  
> I fully admit I appended some print-related event listeners to a javascript file (sdoMap.js) that I didn't otherwise spend the time to totally understand. I'm trusting the baseline knowledge of reviewers for this repo to tell me if that can/should go somewhere else.

### DIFF
--- a/test/baselines/generated-reference/assets-inline.html
+++ b/test/baselines/generated-reference/assets-inline.html
@@ -82,6 +82,18 @@ document.addEventListener('DOMContentLoaded', () => {
   );
 });
 
+window.addEventListener('beforeprint', () => {
+  if (!document.body.classList.contains('oldtoc')) {
+    document.getElementById('metadata-block').before(document.getElementById('menu'));
+  }
+});
+
+window.addEventListener('afterprint', () => {
+  if (!document.body.classList.contains('oldtoc')) {
+    document.getElementById('spec-container').before(document.getElementById('menu'));
+  }
+});
+
 'use strict';
 function Search(menu) {
   this.menu = menu;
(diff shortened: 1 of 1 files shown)


js/sdoMap.js
```diff
@@ -80,3 +80,15 @@ document.addEventListener('DOMContentLoaded', () => {
...
+
+window.addEventListener('beforeprint', () => {
+  if (!document.body.classList.contains('oldtoc')) {
+    document.getElementById('metadata-block').before(document.getElementById('menu'));
+  }
+});
+
+window.addEventListener('afterprint', () => {
+  if (!document.body.classList.contains('oldtoc')) {
+    document.getElementById('spec-container').before(document.getElementById('menu'));
```
bakkot, line 92: This doesn't look like the inverse of `beforeprint`, which it seems like it should be. What's this trying to accomplish?

michaelficarra: I don't think these event handlers operate as inverses of each other as you claim.

gesa:
It improves the readability of the document for visitors wishing to generate paged media. It's an accessibility improvement.

michaelficarra:
I get that, but the question was how do these styles and the moved around DOM nodes help the readability?

js/sdoMap.js
michaelficarra: We're not talking about the `beforeprint` and `afterprint` events. We're saying that running the function you've associated with `beforeprint` and then running the function you've associated with `afterprint` does not return the DOM to the state it was in at the start. These functions are making incorrect assumptions about the initial ordering of the sections in the document, which is (simplified)

```html
<body>
  ...
  <div id="menu">...</div>
  <div id="spec-container">
    ...
    <div id="metadata-block">...</div>
    ...
  </div>
  ...
</body>
```

Running your `beforeprint` and `afterprint` handlers will result in a document like (simplified)

```html
  ...
  <div id="metadata-block">...</div>
  <div id="spec-container">...</div>
  <div id="menu">...</div>
  ...
</body>
```

That's not the initial structure. Possibly worse, it seems the document would have these nodes in a different order during the second print than they were during the first print.

css/print.css
```diff
@@ -186,12 +186,29 @@
   page-break-after: always;
 }
 
+/* Styles for web users who wish to print (or "print") the document */
```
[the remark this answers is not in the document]
michaelficarra: To me, that means print-to-PDF, and it'd be easier for me to understand by calling it that. If there's other "print" functions, we could list them as examples too. It'd give me something concrete to think about when reading this.

js/sdoMap.js
michaelficarra: Oh, whoops, it looks like we had the responsibilities of the `this` value and the parameter swapped. Thanks for the explanation.

PR #656, opened 2025-09-12
Limiting max levels to 5

### LINKED ISSUE #655: Limit clause numbers to five levels
ECMA documents require section numbers to be a maximum of 5 levels deep. While this isn't a challenge for manually authored specs, it may become an issue with spec.html that is automatically generated from other source material.

Expected behavior:
- clauseNums.js should limit the levels to a maximum of 5
- attemps to add new child clauses should not result in a new level
  - rather, continue to increment the fifth level.

stevespringett:
Closes #655

### DIFF
--- a/test/baselines/generated-reference/max-clause-depth.html
+++ b/test/baselines/generated-reference/max-clause-depth.html
@@ -16,7 +16,7 @@
 </ul></div><div id="menu-toggle"><svg xmlns="http://www.w3.org/2000/svg" style="width:100%; height:100%; stroke:currentColor" viewBox="0 0 120 120" width="54" height="54">
       <title>Menu</title>
       <path stroke-width="10" stroke-linecap="round" d="M30,60 h60  M30,30 m0,5 h60  M30,90 m0,-5 h60"></path>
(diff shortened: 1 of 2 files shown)


bakkot:
Thanks for the contribution. This seems like something which should be opt-in, and also should generate a warning since you'll often want manual remediation. I've pushed up commits doing both (+ a test). If this still works for you I'll land.

stevespringett:
@bakkot opt-in sounds reasonable as does displaying a warning.

test/baselines/generated-reference/max-clause-depth.html
```diff
@@ -0,0 +1,39 @@
...
+
+<emu-clause id="one">
+  <h1><span class="secnum">1</span> One</h1>
+  <emu-clause id="two">
+    <h1><span class="secnum">1.1</span> Two</h1>
+    <emu-clause id="three">
+      <h1><span class="secnum">1.2</span> Three</h1>
```
michaelficarra, line 27: I would put a depth even deeper than 3 to make sure the flattening happens recursively.

michaelficarra:
> I don't see why it would be? From what I understand the ISO standards only specify a limit on the numbers themselves, not on the presentation of clauses.

The clause numbering should always coincide with the nesting of `emu-clause`s.

bakkot:
This PR does not do any auto-collapsing. (I originally thought it did but I was misreading.) It only touches the numbering.

> The clause numbering should always coincide with the nesting of `emu-clause`s.

That's a valid preference but unless you intend to use this feature I would prefer to defer to the preferences of the people who are actually going to use this. The warning is still emitted and enforceable with `--strict`, it just also makes the numbering compliant with ISO for people who are not enforcing the warning.

bakkot:
Version 21.4.0 is out including the new `--max-clause-depth` flag.

PR #657, opened 2025-09-15
add --max-clause-depth

bakkot:
Fixes https://github.com/tc39/ecmarkup/issues/655, cc @stevespringett.


bakkot:
Hm, actually, on reading #655 again I think the approach of just incrementing the last number on overflow is reasonable. I'll push these commits to that branch.

PR #663, opened 2025-11-13
Add ToC & max clause depth to provide flexibility

gesa:
Ecma house style is a ToC 3 deep and max clauses 6, but there may be scenarios where a TC wants to generate a draft with fewer or more in the context of their specific standard

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -55,7 +55,7 @@ markEffects: true
       <tr><td>`--assets-dir`</td><td>`assetsDir`</td><td>Directory in which to place assets when using `--assets=external`. Defaults to "assets".</td></tr>
       <tr><td>`--lint-spec`</td><td>`lintSpec`</td><td>Enforce some style and correctness checks.</td></tr>
       <tr><td>`--error-formatter`</td><td></td><td>The <a href="https://eslint.org/docs/user-guide/formatters/">eslint formatter</a> to be used for printing warnings and errors when using `--verbose`. Either the name of a built-in eslint formatter or the package name of an installed eslint compatible formatter.</td></tr>
-      <tr><td>`--max-clause-depth N`</td><td>`maxClauseDepth`</td><td>Warn when clauses exceed a nesting depth of N, and cause those clauses to be numbered by incrementing their parent clause's number rather than by nesting a new number within their parent clause. Default 6.</td></tr>
+      <tr><td>`--max-clause-depth N`</td><td>`maxClauseDepth`</td><td>Warn when clauses exceed a nesting depth of N, and cause those clauses to be numbered by incrementing their parent clause's number rather than by nesting a new number within their parent clause. Default 5.</td></tr>
       <tr><td>`--printed-toc-depth N`</td><td>`printedTocDepth`</td><td>Maximum level of subclause to be emitted in printed table of contents. Default 3. (ignored if `--printable` option omitted)</td></tr>
       <tr><td>`--strict`</td><td></td><td>Exit with an error if there are warnings. Cannot be used with `--watch`.</td></tr>
       <tr><td>`--multipage`</td><td></td><td>Emit a distinct page for each top-level clause.</td></tr>
--- a/src/args.ts
+++ b/src/args.ts
@@ -71,7 +71,7 @@ export const options = [
     name: 'max-clause-depth',
     type: Number,
     description:
-      'The maximum nesting depth for clauses; exceeding this will cause a warning. Defaults to six (per Ecma house style.)',
+      'The maximum nesting depth for clauses; exceeding this will cause a warning. Defaults to five (per Ecma house style.)',
   },
   {
     name: 'printed-toc-depth',


src/clauseNums.ts
```diff
@@ -10,7 +10,8 @@ export default function iterator(spec: Spec): ClauseNumberIterator {
   let inAnnex = false;
   let currentLevel = 0;
   let hasWarnedForExcessNesting = false;
-  const MAX_LEVELS = spec.opts.maxClauseDepth ?? Infinity;
+  // Ecma house style calls for a maximum of 5 levels of clause division
+  const MAX_LEVELS = spec.opts.maxClauseDepth ?? 6;
```
michaelficarra:
```suggestion
  const MAX_LEVELS = spec.opts.maxClauseDepth ? (spec.opts.maxClauseDepth || Infinity) : 6;
```

This would map `0` to Infinity, which seems reasonable for an option that we accept over the command line.

michaelficarra (approved):

spec/index.html
```diff
@@ -55,7 +55,8 @@ <h1>Options</h1>
...
       <tr><td>`--lint-spec`</td><td>`lintSpec`</td><td>Enforce some style and correctness checks.</td></tr>
       <tr><td>`--error-formatter`</td><td></td><td>The <a href="https://eslint.org/docs/user-guide/formatters/">eslint formatter</a> to be used for printing warnings and errors when using `--verbose`. Either the name of a built-in eslint formatter or the package name of an installed eslint compatible formatter.</td></tr>
-      <tr><td>`--max-clause-depth N`</td><td></td><td>Warn when clauses exceed a nesting depth of N, and cause those clauses to be numbered by incrementing their parent clause's number rather than by nesting a new number within their parent clause.</td></tr>
+      <tr><td>`--max-clause-depth N`</td><td>`maxClauseDepth`</td><td>Warn when clauses exceed a nesting depth of N, and cause those clauses to be numbered by incrementing their parent clause's number rather than by nesting a new number within their parent clause. Default 6.</td></tr>
+      <tr><td>`--toc-depth N`</td><td>`tocDepth`</td><td>Maximum level of subclause to be emitted in table of contents. Default 3.</td></tr>
```
bakkot: Looks like this only applies to printed documents? If so this should be named something like `--printable-toc-depth`, or at least have that behavior documented here

gesa: Thank you, parameter and documentation both updated.

gesa:
I received clarification on conflicting direction from the secretariat. The maximum total clause depth is 5 (1.2.3.4.5), _not_ 5 subdivisions. I have updated the PR in https://github.com/tc39/ecmarkup/pull/663/commits/7c88604659de82f77f77d00f69780439902bbdeb

spec/index.html
```diff
@@ -55,7 +55,8 @@ <h1>Options</h1>
       <tr><td>`--assets-dir`</td><td>`assetsDir`</td><td>Directory in which to place assets when using `--assets=external`. Defaults to "assets".</td></tr>
       <tr><td>`--lint-spec`</td><td>`lintSpec`</td><td>Enforce some style and correctness checks.</td></tr>
       <tr><td>`--error-formatter`</td><td></td><td>The <a href="https://eslint.org/docs/user-guide/formatters/">eslint formatter</a> to be used for printing warnings and errors when using `--verbose`. Either the name of a built-in eslint formatter or the package name of an installed eslint compatible formatter.</td></tr>
-      <tr><td>`--max-clause-depth N`</td><td></td><td>Warn when clauses exceed a nesting depth of N, and cause those clauses to be numbered by incrementing their parent clause's number rather than by nesting a new number within their parent clause.</td></tr>
+      <tr><td>`--max-clause-depth N`</td><td>`maxClauseDepth`</td><td>Warn when clauses exceed a nesting depth of N, and cause those clauses to be numbered by incrementing their parent clause's number rather than by nesting a new number within their parent clause. Default 6.</td></tr>
```
michaelficarra:
```suggestion
      <tr><td>`--max-clause-depth N`</td><td>`maxClauseDepth`</td><td>Warn when clauses exceed a nesting depth of N, and cause those clauses to be numbered by incrementing their parent clause's number rather than by nesting a new number within their parent clause. Default 5.</td></tr>
```

src/clauseNums.ts
```diff
@@ -10,7 +10,8 @@ export default function iterator(spec: Spec): ClauseNumberIterator {
   let inAnnex = false;
   let currentLevel = 0;
   let hasWarnedForExcessNesting = false;
-  const MAX_LEVELS = spec.opts.maxClauseDepth ?? Infinity;
+  // Ecma house style calls for a maximum of 5 clause levels
+  const MAX_LEVELS = spec.opts.maxClauseDepth ? spec.opts.maxClauseDepth || Infinity : 5;
```
gibson042: Defaulting to 5 results in errors when building ECMA-262... I'd rather keep the infinite default and encode this default in build scripts.
```suggestion
  const MAX_LEVELS = spec.opts.maxClauseDepth || Infinity;
```

src/args.ts
```diff
@@ -76,7 +76,13 @@ export const options = [
     name: 'max-clause-depth',
     type: Number,
     description:
-      'The maximum nesting depth for clauses; exceeding this will cause a warning. Defaults to no limit.',
+      'The maximum nesting depth for clauses; exceeding this will cause a warning. Defaults to five (per Ecma house style.)',
```
gibson042, line 79:
```suggestion
      'The maximum nesting depth for clauses; exceeding this will cause a warning. Defaults to no limit; Ecma house style is 5.',
```

spec/index.html
```diff
@@ -55,7 +55,8 @@ <h1>Options</h1>
       <tr><td>`--assets-dir`</td><td>`assetsDir`</td><td>Directory in which to place assets when using `--assets=external`. Defaults to “assets”.</td></tr>
       <tr><td>`--lint-spec`</td><td>`lintSpec`</td><td>Enforce some style and correctness checks.</td></tr>
       <tr><td>`--error-formatter`</td><td></td><td>The <a href="https://eslint.org/docs/user-guide/formatters/">eslint formatter</a> to be used for printing warnings and errors when using `--verbose`. Either the name of a built-in eslint formatter or the package name of an installed eslint compatible formatter.</td></tr>
-      <tr><td>`--max-clause-depth N`</td><td></td><td>Warn when clauses exceed a nesting depth of N, and cause those clauses to be numbered by incrementing their parent clause's number rather than by nesting a new number within their parent clause.</td></tr>
+      <tr><td>`--max-clause-depth N`</td><td>`maxClauseDepth`</td><td>Warn when clauses exceed a nesting depth of N, and cause those clauses to be numbered by incrementing their parent clause's number rather than by nesting a new number within their parent clause. Default 5.</td></tr>
```
gibson042:
```suggestion
      <tr><td>`--max-clause-depth N`</td><td>`maxClauseDepth`</td><td>Warn when clauses exceed a nesting depth of N, and cause those clauses to be numbered by incrementing their parent clause's number rather than by nesting a new number within their parent clause. Ecma house style is 5.</td></tr>
```

src/cli.ts
```diff
@@ -116,6 +116,9 @@ const build = debounce(async function build() {
     if (args['max-clause-depth']) {
       opts.maxClauseDepth = args['max-clause-depth'];
     }
+    if (args['printed-toc-depth']) {
```
gibson042: CLI <kbd>--max-clause-depth=0</kbd> results in `args['printed-toc-depth']` being the number 0, which is falsy but presumably should still be propagated.
```suggestion
    if (args['printed-toc-depth'] != null) {
```

gibson042 (changes requested):
Is it intentional to introduce errors when building the ECMA-262 draft?

src/cli.ts
michaelficarra: Should it just be required to be positive if provided?

gibson042: It could be, but I find your [argument for interpreting `--max-clause-depth=0` as "no limit"](https://github.com/tc39/ecmarkup/pull/663#discussion_r2521519524) to be persuasive.

gesa:
> Is it intentional to introduce errors when building the ECMA-262 draft?

@gibson042 Yes and no. Ecma house style limits clause depth to 5. ECMA-262's clauses exceed that. It's _fine_, in that we'll leave ECMA-262 specifically like that after 17 editions. spec.html can add `maxClauseDepth: 6` (or even `maxClauseDepth: 0`, but I have a strong preference for the former) to its metadata. But any other standards that TC39 publishes which don't already have the legacy baggage should stick to the house style, and ECMA-262 should avoid subclauses any deeper than it already has.

As far as defaults go, I'm a fan of 0 == Infinity, unset == 5 for clause depth, unset == 3 for ToC depth. Also < 0 setting for ToC just omits it altogether.

src/cli.ts
```diff
@@ -116,6 +116,9 @@ const build = debounce(async function build() {
     if (args['max-clause-depth']) {
```
gibson042:
```suggestion
    if (args['max-clause-depth'] != null) {
```

gibson042 (approved):
> [ECMA-262] spec.html can add `maxClauseDepth: 6` (or even `maxClauseDepth: 0`, but I have a strong preference for the former) to its metadata. But any other standards that TC39 publishes which don't already have the legacy baggage should stick to the house style, and ECMA-262 should avoid subclauses any deeper than it already has.

Thanks, that clarifies expectations for how to continue running `ecmarkup` with `--strict` in ECMA-262.

This PR looks fine to me after fixing cli.ts.

PR #664, opened 2025-11-13
Table and Figure caption styles

gesa:
Updated to Ecma house style.

- Update css to align with template
- Replace `: ` with ` — `
- Begin re-numbering figures at the beginning of each annex (as A1, A2, etc)

### DIFF
--- a/src/Figure.ts
+++ b/src/Figure.ts
@@ -56,6 +56,11 @@ export default class Figure extends Builder {
 
     const captionElem = spec.doc.createElement('figcaption');
     captionElem.innerHTML = figure.caption;
+
+    if (node.nodeName.split('-')[1] === 'FIGURE') {
+      node.firstElementChild?.appendChild(captionElem);
+    } else {
       node.childNodes[0].insertBefore(captionElem, node.childNodes[0].firstChild);
+    }
   }
 }
--- a/test/baselines/generated-reference/figure.html
+++ b/test/baselines/generated-reference/figure.html
@@ -11,14 +11,14 @@
   <li><span>Jump to the most recent link target</span><code>`</code></li>
 </ul></div><div id="spec-container">
 <link rel="stylesheet" href="css/elements.css">
-<emu-figure id="figure-1"><figure><figcaption>Figure 1</figcaption>
+<emu-figure id="figure-1"><figure>
   this is a figure!
-</figure></emu-figure>
+<figcaption>Figure 1</figcaption></figure></emu-figure>
 
 <!-- deprecated: caption element -->
-<emu-figure id="figure-2" informative="" caption="Informative figure"><figure><figcaption>Figure 2 (Informative) — Informative figure</figcaption>
+<emu-figure id="figure-2" informative="" caption="Informative figure"><figure>
   this is a figure!
-</figure></emu-figure>
+<figcaption>Figure 2 (Informative) — Informative figure</figcaption></figure></emu-figure>
 
 <emu-table id="table-1" caption="An example table"><figure><figcaption>Table 1 — An example table</figcaption>
   <table>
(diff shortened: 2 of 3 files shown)


src/Figure.ts
```diff
@@ -21,16 +21,16 @@ export default class Figure extends Builder {
...
-    this.caption = this.type.charAt(0).toUpperCase() + this.type.slice(1) + ' ' + this.number;
+    this.caption = `${this.type.charAt(0).toUpperCase()}${this.type.slice(1)} ${spec._figurePrefix}${this.number}`;
 
     if (this.isInformative) {
       this.caption += ' (Informative)';
     }
 
     if (this.captionElem) {
-      this.caption += ': ' + this.captionElem.innerHTML;
+      this.caption += ' — ' + this.captionElem.innerHTML;
```
michaelficarra:
```suggestion
      this.caption += ' \u2014 ' + this.captionElem.innerHTML;
```

To make it a little more clear that this isn't an ordinary hyphen. Same below.

michaelficarra (approved):
I don't like it, but okay.

gesa:
> I don't like it, but okay.

@michaelficarra I would be delighted to come at the numbering with a different approach. I also don't like it, at least not the solution I'm proposing. I just couldn't think of anything better.

I found a mismatch in the layout as compared to the template reviewing this—figure captions should render _below_ the figure, table captions should render _above_ the table. Commit resolving this incoming.

src/Figure.ts
```diff
@@ -60,6 +60,11 @@ export default class Figure extends Builder {
 
     const captionElem = spec.doc.createElement('figcaption');
     captionElem.innerHTML = figure.caption;
-    node.childNodes[0].insertBefore(captionElem, node.childNodes[0].firstChild);
+
+    if (node.nodeName.split('-')[1] === 'FIGURE') {
```
gibson042:
```suggestion
    if (node.nodeName === 'EMU-FIGURE') {
```

src/Figure.ts
```diff
@@ -60,6 +60,11 @@ export default class Figure extends Builder {
...
+      node.firstElementChild?.appendChild(captionElem);
+    } else {
+      node.childNodes[0].insertBefore(captionElem, node.childNodes[0].firstChild);
+    }
```
gibson042:
```suggestion
    // Captions go below figures but above tables etc. (per Ecma house style)
    if (node.nodeName === 'EMU-FIGURE') {
      ele.appendChild(captionElem);
    } else {
      ele.insertBefore(captionElem, ele.firstChild);
    }
```

src/Clause.ts
```diff
@@ -111,6 +111,13 @@ export default class Clause extends Builder {
       parentNamespace = parent.namespace;
     }
 
+    // Reset figure numbering for each Annex
+    if (this.isAnnex && !this.parentClause) {
+      spec._figureCounts.figure = 0;
+      spec._figureCounts.table = 0;
+      spec._figurePrefix = `${number}.`;
+    }
```
gibson042, line 119: This breaks the rendering of \<emu-xref>, e.g. "_The entries in Table 103 are added to Table 6_" at [B.2.1 Additional Properties of the Global Object](https://tc39.es/ecma262/multipage/additional-ecmascript-features-for-web-browsers.html#sec-additional-properties-of-the-global-object) becomes "_The entries in Table 1 are added to Table 6_" rather than "_The entries in Table B.1 are added to Table 6_".

gibson042 (changes requested):

src/Clause.ts
gesa: This seemed like a suspiciously easy fix. The figure biblio entries expect a number _or_ a string, so I modified the prefix to be added at that point in the figure generation.

Issue #391, opened 2021-12-13
better support for concrete methods


bakkot:
I feel like this is already an open issue, but I couldn't find it, so making it again I guess.

We need to make concrete methods linkable and have some way to enforce that their signatures line up. Probably this means having a central place where the main definition lives, with (automatically generated) links to the various implementations.

bakkot:
Maybe internal methods, too.

jmdyck:
> Probably this means having a central place where the main definition lives

This already exists. E.g., for the 3 `HasBinding` concrete methods, the 'central place' is in [Table 16](https://tc39.es/ecma262/#table-abstract-methods-of-environment-records). For enforcing that signatures line up, the table declaration gives the number and names of the parameters, but not their types (or the return type). Or rather, that type info is mostly there in the "Purpose" column, but not in a way that can be easily extracted by code. 

So are you thinking that the tabular forms would be replaced with structured-header forms?

PR #666, opened 2026-01-16
Improve linking for abstract&concrete methods

nicolo-ribaudo:
Ref #391

The first commit:
- adds support for structured headers in abstract method declarations. A list of abstract method declarations is introduced by `<emu-table type="abstract methods">`.
- adds abstract methods as `op`s to the biblio, so that autolinking will find their references and auto-link to the declaration

The second commit:
- adds concrete methods to the biblio
- introduces an `<emu-concrete-method-dfns for="SomeAbstractMethodName">`, which auto-expands to a list of definitions for `SomeAbstractMethodName`

With both together, this code:
```html
  <emu-table type="abstract methods" of="Module Record" id="table-abstract-methods-of-module-records">
    <table>
      <thead>
        <tr>
          <th>Method</th>
          <th>Description</th>
          <th>Definitions</th>
        </tr>
      </thead>
      <tr>
        <td>
          ResolveExport (
            _exportName_: a String,
            <ins>_resolveSet_: a List of Module Records,</ins>
          ): either a Module Record or *null*
        </td>
        <td>It does something when called.</td>
        <td>
          It is defined in the following places:
          <emu-concrete-method-dfns for="ResolveExport"></emu-concrete-method-dfns>
        </td>
      </tr>
    </table>
  </emu-table>
```

Is rendered like this:

image

This PR does not introduce yet any checks that the signatures match.

Some questions:
1. Right now the links from usage sides to abstract method definitions just link to the table, should they have more precise links?
2. Usually AO signatures are wrapped in `<h1>`, but I'm not sure it makes much sense to do it inside a table. Should I do it anyway for consistency?

### DIFF
--- a/src/type-logic.ts
+++ b/src/type-logic.ts
@@ -608,9 +608,6 @@ export function typeFromExprType(type: BiblioType): Type {
       if (text === '*undefined*') {
         return { kind: 'undefined' };
       }
-      if (text === '*undefined*') {
-        return { kind: 'undefined' };
-      }
       break;
     }
     case 'unused': {


bakkot:
This looks great, thank you!

Can you also add a test which confirms that we now get signature-based errors for invocations of abstract methods? E.g. that if we have an abstract method `Method` with signature `( _x_: a string )`, then invoking it as `_foo_.Method(~enum~)` and `_foo_.Method(*"string"*, *"extra"*)` both give warnings.

src/ConcreteMethodDfns.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    const definitions = spec.biblio.byAbstractMethodAoid(this._for, namespace) ?? [];
+
+    const ul = spec.doc.createElement('ul');
+
+    for (const def of definitions) {
+      const li = spec.doc.createElement('li');
+      const xrefNode = spec.doc.createElement('emu-xref');
+      xrefNode.setAttribute('href', `#${def.refId}`);
+      xrefNode.textContent = def.for;
```
michaelficarra: Can we prefix it with the section number?

bakkot: I think it should be possible to get that for free by re-use the existing emu-xref machinery by simply not populating textContent here. Alternatively, if that doesn't work because that phase has already run or something, we can [export and re-use the relevant function directly](https://github.com/tc39/ecmarkup/blob/44d8f38317086e7ffede4dc704979123b5b7d3de/src/Xref.ts#L391).

nicolo-ribaudo: Not putting the text content seems to give _just_ the number, without the text. I'll do it manually.

Would we want the number of the section (e.g. `16.2.1.2 Source Text Module Record`) or of the specific concrete method that we are linking to (e.g. `16.2.1.7.2.2 Source Text Module Record`)?

nicolo-ribaudo:
> invoking it as `_foo_.Method(~enum~)`

Is this meant to warn because of the mismatched argument type? I do not seem to be getting that warning even with plain AOs.

src/ConcreteMethodDfns.ts
nicolo-ribaudo: I'm going ahead with the full number, since that seems to be more useful when manually looking for the number (e.g. in a printed version), even if the title doesn't match.

bakkot:
> Is this meant to warn because of the mismatched argument type? I do not seem to be getting that warning even with plain AOs.

Sorry, has to be spelled "a String", with the capital letter, to actually work. Full example:

```html
<pre class="metadata">
copyright: false
</pre>
<emu-clause id="example" type="abstract operation">
  <h1>
    Example (
      arg: a String,
    ): ~unused~
  </h1>
  <dl class="header"></dl>
</emu-clause>
<emu-alg>
  1. Perform Example(~enum~).
</emu-alg>
```

```
$ ./bin/ecmarkup.js --lint-spec test.emu /dev/null
Warning: inlining assets; this should not be done for production builds.
[2026-01-16T21:03:31.921Z] Warning: test.emu:13:22: argument (~enum~) does not look plausibly assignable to parameter type (String)
```

nicolo-ribaudo:
Oh yes that worked, thanks.

michaelficarra (approved):
🆒

src/type-logic.ts
```diff
@@ -584,6 +608,9 @@ export function typeFromExprType(type: BiblioType): Type {
       if (text === '*undefined*') {
         return { kind: 'undefined' };
       }
+      if (text === '*undefined*') {
+        return { kind: 'undefined' };
+      }
```
michaelficarra: huh?

PR #678, opened 2026-03-14
Terms & Defs style rule

gesa:
- Not every Ecma standard prepends their claueses with `sec`, this should catch the rest.
- terms are 10pt font, not styled like section headers.

### DIFF
--- a/css/elements.css
+++ b/css/elements.css
@@ -256,12 +256,12 @@
 
 html {
   background-color: var(--background-color);
+  font-size: 18px;
 }
 
 body {
   display: flex;
   word-wrap: break-word;
-  font-size: 18px;
   line-height: 1.5;
   font-family:
     'IBM Plex Serif',
@@ -918,6 +918,18 @@ emu-annex emu-annex emu-annex emu-annex emu-annex emu-annex {
 #spec-container > emu-clause:first-of-type,
 #spec-container > emu-annex:first-of-type {
   margin-top: 0;
+}
+
+[id*="terms-and-definitions"] emu-clause h1 {
+  font-size: 1rem;
+}
+
+[id*="terms-and-definitions"] emu-clause h1 + p {
+  margin-top: 0;
+}
+
+[id*="terms-and-definitions"] emu-clause .secnum {
+  display: block;
 }
 
 /* Figures and tables */
--- a/css/elements.css
+++ b/css/elements.css
@@ -920,15 +920,15 @@ emu-annex emu-annex emu-annex emu-annex emu-annex emu-annex {
   margin-top: 0;
 }
 
-[id*="terms-and-definitions"] emu-clause h1 {
+[id*='terms-and-definitions'] emu-clause h1 {
   font-size: 1rem;
 }
 
-[id*="terms-and-definitions"] emu-clause h1 + p {
+[id*='terms-and-definitions'] emu-clause h1 + p {
   margin-top: 0;
 }
 
-[id*="terms-and-definitions"] emu-clause .secnum {
+[id*='terms-and-definitions'] emu-clause .secnum {
   display: block;
 }
 
--- a/css/print.css
+++ b/css/print.css
@@ -660,19 +660,24 @@ p.ECMAaddress {
   margin: 0;
 }
 
-#sec-terms-and-definitions dfn {
+[id$=terms-and-definitions] dfn {
   font-style: normal;
 }
 
-#sec-terms-and-definitions h1 .secnum {
+[id$=terms-and-definitions] emu-clause h1 {
+  font-size: 10pt;
+  -prince-bookmark-level: none;
+}
+
+[id$=terms-and-definitions] h1 .secnum {
   display: block;
 }
 
-#sec-terms-and-definitions > h1 > .secnum {
+[id$=terms-and-definitions] > h1 > .secnum {
   display: inline;
 }
 
-#sec-terms-and-definitions h1 + p {
+[id$=terms-and-definitions] h1 + p {
   margin-top: 0;
 }
 
--- a/css/print.css
+++ b/css/print.css
@@ -188,11 +188,11 @@
 
 html, body {
   background-color: initial;
+  font-size: 10pt;
 }
 
 body {
   font-family: 'Arial Plus', Arial, Helvetica, sans-serif, "DejaVu Math TeX Gyre", Symbola, monospace;
-  font-size: 10pt;
   color: #000;
   line-height: 1.15;
 }
@@ -660,25 +660,13 @@ p.ECMAaddress {
   margin: 0;
 }
 
-[id$=terms-and-definitions] dfn {
+[id*=terms-and-definitions] dfn {
   font-style: normal;
 }
 
-[id$=terms-and-definitions] emu-clause h1 {
+[id*=terms-and-definitions] emu-clause h1 {
   font-size: 10pt;
   -prince-bookmark-level: none;
-}
-
-[id$=terms-and-definitions] h1 .secnum {
-  display: block;
-}
-
-[id$=terms-and-definitions] > h1 > .secnum {
-  display: inline;
-}
-
-[id$=terms-and-definitions] h1 + p {
-  margin-top: 0;
 }
 
 p.adoption-info {
--- a/css/print.css
+++ b/css/print.css
@@ -660,12 +660,11 @@ p.ECMAaddress {
   margin: 0;
 }
 
-[id*=terms-and-definitions] dfn {
+[id*='terms-and-definitions'] dfn {
   font-style: normal;
 }
 
-[id*=terms-and-definitions] emu-clause h1 {
-  font-size: 10pt;
+[id*='terms-and-definitions'] emu-clause h1 {
   -prince-bookmark-level: none;
 }
 
--- a/js/print.js
+++ b/js/print.js
@@ -30,7 +30,7 @@ PDF.subject = shortname.innerHTML + (version ? ', ' + version.innerHTML : '');
 /**
  * Terms and definitions section should not have every term listed in the table of contents.
  * */
-const terms = document.querySelector('#toc a[href*="terms-and-definitions"]');
+const terms = document.querySelector('#toc a[href*="#terms-and-definitions"]');
 
 if (terms) {
   (terms.parentElement.querySelector('ol.toc') || document.createElement('i')).remove();
--- a/js/print.js
+++ b/js/print.js
@@ -30,7 +30,7 @@ PDF.subject = shortname.innerHTML + (version ? ', ' + version.innerHTML : '');
 /**
  * Terms and definitions section should not have every term listed in the table of contents.
  * */
-const terms = document.querySelector('#toc a[href="#sec-terms-and-definitions"]');
+const terms = document.querySelector('#toc a[href*="terms-and-definitions"]');
 
 if (terms) {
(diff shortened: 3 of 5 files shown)


css/print.css
```diff
@@ -660,19 +660,24 @@ p.ECMAaddress {
   margin: 0;
 }
 
-#sec-terms-and-definitions dfn {
+[id$=terms-and-definitions] dfn {
```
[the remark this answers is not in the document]
gesa: @gibson042 Thanks for this comment, it was a good reminder to move some of my document prep functionality into the ecmarkup repo. The terms & defs clause is unique, and the ID is used to manipulate the document beyond styles.

Resolving this and I'll ask that if you still feel we should be using classes instead of IDs, you start a new review. My apologies for the time you already put into reviewing this PR.

js/print.js
```diff
@@ -30,7 +30,7 @@ PDF.subject = shortname.innerHTML + (version ? ', ' + version.innerHTML : '');
 /**
  * Terms and definitions section should not have every term listed in the table of contents.
  * */
-const terms = document.querySelector('#toc a[href="#sec-terms-and-definitions"]');
+const terms = document.querySelector('#toc a[href*="#terms-and-definitions"]');
```
gesa, line 33:
```suggestion
const terms = document.querySelector('#toc a[href*="terms-and-definitions"]');
```
🤦🏻‍♀️

spec/index.html
```diff
@@ -164,7 +164,9 @@ <h2>Example</h2>
 
 <emu-clause id="clauses">
   <h1>Clauses</h1>
-  <p>Clauses are referenced using their id and are numbered automatically based on document position. Ecmarkdown syntax can be used in descendent text nodes as well. Text nodes are parsed as Ecmarkdown Fragments.</p>
+  <p>Clauses are referenced using their id and are numbered automatically based on document position. Ecmarkdown syntax can be used in descendent text nodes as well. Text nodes are parsed as Ecmarkdown Fragments. Clauses always begin with an `&lt;h1>` element.</p>
```
gibson042, line 167:
```suggestion
  <p>Clauses are referenced using their id and are numbered automatically based on document position. Ecmarkdown syntax can be used in descendent text nodes as well. Text nodes are parsed as Ecmarkdown Fragments. Clauses must always begin with an `&lt;h1>` element.</p>
```

css/elements.css
```diff
@@ -920,6 +920,18 @@ emu-annex emu-annex emu-annex emu-annex emu-annex emu-annex {
...
+  font-size: 1rem;
+}
+
+[id*='terms-and-definitions'] emu-clause h1 + p {
+  margin-top: 0;
+}
+
+[id*='terms-and-definitions'] emu-clause .secnum {
+  display: block;
+}
```
gibson042, line 933: I would still prefer matching on class rather than id-contains-substring.
```suggestion
.ecma-terms-and-definitions emu-clause h1 {
  font-size: 1rem;
}

.ecma-terms-and-definitions emu-clause h1 + p {
  margin-top: 0;
}

.ecma-terms-and-definitions emu-clause .secnum {
  display: block;
}
```
(and likewise in css/print.css)

spec/index.html
```diff
@@ -255,6 +256,42 @@ <h1>
         <p>This clause is normative optional.</p>
       </emu-clause>
     </aside>
+
+    <emu-clause id="terms">
+      <h1>Terms and definitions</h1>
+      <p>Ecmarkup produces special formatting behaviour for an `&lt;emu-clause>` with an ID ending in `terms-and-definitions`. No action is required on an editor's part to properly format that clause as long as it uses similar markup to any other clause, i.e. an `&lt;h1>` element followed by one or more `&lt;p>` or `&lt;emu-clause>` elements.</p>
```
gibson042, line 262: The selectors in elements.css and print.css use `*=`, which match anywhere in the attribute value rather than just at the end (which would be `$=`). If this PR were to land now, it would affect not just [ECMA-262 Terms and Definitions](https://tc39.es/ecma262/multipage/overview.html#sec-terms-and-definitions) but also its subclauses.

spec/index.html
```diff
@@ -255,6 +256,42 @@ <h1>
...
+      <p>Using a `&lt;dfn>` element in the clause title is optional, and will allow ecmarkup to link back to the term anywhere it is used in the rest of the document. See <emu-xref href="#definitions">definitions</emu-xref>.</p>
```
gibson042, line 263: Use of \<dfn> in \<h1> content _works_, but seems like poor form. I'd rather not encourage it.

spec/index.html
```diff
@@ -1104,3 +1141,47 @@ <h1>Code Listings</h1>
...
+    <p>Ecma standards conforming to Ecma house style begin with an unnumbered introduction clause (using a single `&lt;emu-intro>` element), followed by the following numbered clauses:</p>
+
+    <ol>
+      <li>Scope</li>
+      <li>Normative references</li>
+      <li>Informative references</li>
+      <li>Terms and definitions</li>
+    </ol>
+
+    <p>(Refer to the Ecma house style guide for more detailed information.)</p>
```
gibson042, line 1185: How might an arbitrary reader access the Ecma house style guide?

spec/index.html
```diff
@@ -255,6 +256,42 @@ <h1>
...
+            <h1>technical definition</h1>
+            <p>descriptive text aligning with Ecma house style, providing text that could be swapped in for the term without impacting readability</p>
+          </emu-clause>
+
+          <emu-clause id="term-ecma">
+            <h1><dfn>Ecma International</dfn></h1>
+            <p>international standards organization responsible for technology-related standards since 1959</p>
+          </emu-clause>
+        </emu-clause>
+      </aside>
```
gibson042, line 293: Documenting my suggestion here as well:
```suggestion
      <p>Ecmarkup produces special formatting behaviour for a clause with class `ecma-terms-and-definitions`, of which there should be at most one (and typically should also have id "sec-terms-and-definitions" or "terms-and-definitions"). No action is required on an editor's part to properly format that clause as long as it uses similar markup to any other clause, i.e. an `&lt;h1>` element followed by one or more `&lt;p>` or `&lt;emu-clause>` elements.</p>
      <p>Using a `&lt;dfn>` element in the clause title is optional, and will allow ecmarkup to link back to the term anywhere it is used in the rest of the document. See <emu-xref href="#definitions">definitions</emu-xref>.</p>

      <h2>Example</h2>
      <pre><code class="language-html">
      &lt;emu-clause id="sec-terms-and-definitions" class="ecma-terms-and-definitions">
        &lt;h1>Terms and definitions&lt;/h1>
        &lt;emu-clause id="term-technical-definition">
          &lt;h1>technical definition&lt;/h1>
          &lt;p>descriptive text aligning with Ecma house style, providing text that could be swapped in for the term without impacting readability&lt;/p>
        &lt;/emu-clause>
        &lt;emu-clause id="term-ecma">
          &lt;h1>&lt;dfn>Ecma International&lt;/dfn>&lt;/h1>
          &lt;p>international standards organization responsible for technology-related standards since 1959&lt;/p>
        &lt;/emu-clause>
      &lt;/emu-clause></code></pre>

      <h3>Result</h3>
      <aside class="result">
        <emu-clause id="sec-terms-and-definitions" class="ecma-terms-and-definitions">
          <h1>Terms and definitions</h1>
          <emu-clause id="term-technical-definition">
            <h1>technical definition</h1>
            <p>descriptive text aligning with Ecma house style, providing text that could be swapped in for the term without impacting readability</p>
          </emu-clause>

          <emu-clause id="term-ecma">
            <h1><dfn>Ecma International</dfn></h1>
            <p>international standards organization responsible for technology-related standards since 1959</p>
          </emu-clause>
        </emu-clause>
      </aside>
```

gibson042 (commented):
> The terms & defs clause is unique, and the ID is used to manipulate the document beyond styles.

Can you elaborate on the nature of that manipulation and why it cannot also use a class selector? Or alternatively, I'd be fine with using selectors that rely upon exact matching against a list of `id` values, e.g. `#sec-terms-and-definitions emu-clause h1, #terms-and-definitions emu-clause h1`, although that is still more magical than explicit opt-in via `class`.

gesa:
@gibson042 I was a front-end developer for a long time, I used to have a near-absolute policy against using IDs for styling. I do not like it, I'm not excited about it. But this PR isn't based on being prescriptive, it's descriptive. A half dozen standards already use `#sec-terms-and-definitions` (like ecma262) or `#terms-and-definitions` (like ecma419) for their terms clause. Given that clause is necessarily unique in how it is styled, this is the best way IMO to handle it. Adding a `ecma-terms-and-definitions` class does not solve the problems this PR is addressing, but I'd be thrilled to hear any other options.

n.b. I initially switched from using `[id$=""]` to `[id*=""]` because I found one standard that had appended something to the ID string. I've switched back to using `$` in this PR. I'd also be open to switching to using `#terms-and-definitions, #sec-terms-and-definitions` everywhere and fixing that one standard that would be broken by this whenever it comes up.

I also agree that I don't love using `<dfn>` in `<h1>`, but it seemed like that or writing some amount of logic for ecmarkup to treat `#(sec-)terms-and-definitions emu-clause > h1` as the same thing as `<dfn>` and I don't have the bandwidth to sit down and figure out how easy or hard that would be.

Some examples of how ts&ds is unique:

* subclauses do not appear in table of contents
* subclauses are treated as being numbered but without a title (see below example, but disregard the line-heights)
* each part of a term/definition has specific style rules associated with them

See TOOLS-017.dotx on the Ecma members file server for a visual representation. I'm working on getting that added to the Ecma directives page on the website too.


_example term/definition_

> ## 4 Terms & definitions
> 
> #### 4.1
> **international standard**
> standard that is adopted by an international standardizing/standards organization and made available to the public

gibson042:
> @gibson042 I was a front-end developer for a long time, I used to have a near-absolute policy against using IDs for styling. I do not like it, I'm not excited about it. But this PR isn't based on being prescriptive, it's descriptive. A half dozen standards already use `#sec-terms-and-definitions` (like ecma262) or `#terms-and-definitions` (like ecma419) for their terms clause. Given that clause is necessarily unique in how it is styled, this is the best way IMO to handle it. Adding a `ecma-terms-and-definitions` class does not solve the problems this PR is addressing, but I'd be thrilled to hear any other options.

Can you explain how an `ecma-terms-and-definitions` class would fail to address the special styling needs of Ecma Terms and Definitions sections?

> n.b. I initially switched from using `[id$=""]` to `[id*=""]` because I found one standard that had appended something to the ID string. I've switched back to using `$` in this PR. I'd also be open to switching to using `#terms-and-definitions, #sec-terms-and-definitions` everywhere and fixing that one standard that would be broken by this whenever it comes up.

I could live with `#terms-and-definitions, #sec-terms-and-definitions` selectors, and if that's the direction this is headed then you can disregard the previous request.

> I also agree that I don't love using `<dfn>` in `<h1>`, but it seemed like that or writing some amount of logic for ecmarkup to treat `#(sec-)terms-and-definitions emu-clause > h1` as the same thing as `<dfn>` and I don't have the bandwidth to sit down and figure out how easy or hard that would be.

Honestly, that seems like it would be worse. At least the \<dfn>s are explicit.

spec/index.html
gesa: Updated text with "shall"

gesa:
I agree on preferring explicit `<dfn>` tags. I've updated the css and, joy of joys, the latest edition of prince-books supports `:is()`!

PR #680, opened 2026-03-18
Disallow mutating variables after they are captured

nicolo-ribaudo:
This implements linting for these two new editorial conventions:

> - captured aliases in ACs should not be written to, even if the write is only observed within the AC
> - aliases that are captured by an AC should not be written to outside the AC after their capture

Note that with these conventions we don't catch all cases, e.g.
```
1. Let _curr_ be ~empty~.
2. For _item_ of _list_, do
   a. Set _curr_ to _item_.
   b. Let _ac_ be a new Abstract Closure that captures _curr_ ...
   c. Do something with _curr_.
```

If we want to catch all cases I have a separate patch that implements a more strict set of checks:
> - captured aliases in ACs should not be written to, even if the write is only observed within the AC
> - aliases that are captured by an AC should not be written to outside the AC ~~after their capture~~

the patch is at https://github.com/nicolo-ribaudo/ecmarkup/commit/9d3664d6c51e309ed8a95c3954b81a37b34e50b5, but it would require some tweaks to the spec because we capture reassigned variables 5 times (https://github.com/nicolo-ribaudo/ecma262/commit/d08010afe63bff6f575719d35b5168d13171b5f9).

### DIFF
--- a/src/lint/rules/variable-use-def.ts
+++ b/src/lint/rules/variable-use-def.ts
@@ -41,10 +41,12 @@ type VarKind =
 class Scope {
   declare vars: Map<string, { kind: VarKind; used: boolean; node: HasLocation | null }>;
   declare strictScopes: Set<string>[];
+  declare captured: Set<string>;
   declare report: Reporter;
   constructor(report: Reporter) {
     this.vars = new Map();
     this.strictScopes = [new Set()];
+    this.captured = new Set();
     this.report = report;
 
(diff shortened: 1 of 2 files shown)


bakkot (approved):
Great, thank you!

I personally would be happy with forbidding any assignments to captured variables, even assignments preceding the AC, but this is fine for the moment and I understand it to be all that editors agreed on thus far.

PR #683, opened 2026-03-28
add optional marker for early exits in algorithm steps

### LINKED ISSUE #673: highlight all AO exit points when you click return/throw
An AO can exit from any of its steps via return/throw/`?`. Sometimes I want to see them highlighted in the same way we highlight occurrences of aliases. I think clicking on any return/throw in the algorithm should highlight all exits (including `?`).

michaelficarra:
Fixes #673. I went with a global toggle like we use with user code annotations, rather than per-algorithm.

image

image

image

Disclaimer: I used AI in the process of making this PR.

### DIFF
--- a/css/elements.css
+++ b/css/elements.css
@@ -462,7 +462,8 @@ emu-val {
   font-weight: bold;
 }
 
-emu-alg li.exit:not(emu-alg ol.ac-body > li:last-child):not(emu-alg > ol > li:last-child)::before {
+/* prefix algorithm-terminating steps with an indicator when dynamically enabled */
+emu-alg li.exit::before {
   display: none;
   border: 2pt solid var(--annotation-foreground-color);
   font-size: small;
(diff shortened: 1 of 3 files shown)


css/elements.css
```diff
@@ -462,6 +462,24 @@ emu-val {
...
+  color: var(--annotation-foreground-color);
+  padding: 0 0.25em;
+  background-color: var(--background-color);
+  content: '⏎';
+  margin-right: 0.25em;
+  line-height: 15px;
+  vertical-align: middle;
+}
+
+html.show-early-exits emu-alg li.exit:not(emu-alg ol.ac-body > li:last-child):not(emu-alg > ol > li:last-child)::before {
```
gibson042: We're pretty light on comments in the CSS, but it might be nice to include something like
```css
/* prefix algorithm-terminating steps with an indicator when dynamically enabled */
```

PR #684, opened 2026-03-28
add a visual indicator to AC bodies

### LINKED ISSUE #682: visually distinguish AC bodies
I've been playing around with adding some kind of visual distinction for AC bodies so a reader can easily tell that those steps are not immediately evaluated, even when reading an algorithm from bottom to top.

What do you think about something like this?

Image

It's just a couple styles on the AC's body `<ol>`:

```css
border-left: 3px solid #CCC;
border-radius: 0 0 0 20px;
padding-bottom: 5px;
```

michaelficarra:
Fixes #682.

image

image

Disclaimer: I used AI in the process of making this PR.

### DIFF
--- a/src/minify.ts
+++ b/src/minify.ts
@@ -1,6 +1,6 @@
 import { minify as htmlMinify } from 'html-minifier-terser';
 import * as CleanCSS from 'clean-css';
-import { optimize as svgoOptimize, type Config as SvgoConfig, type PluginConfig } from 'svgo';
+import { optimize as svgoOptimize, type Config as SvgoConfig } from 'svgo';
 import { minify as terserMinify } from 'terser';
 import * as path from 'path';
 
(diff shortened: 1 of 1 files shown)


src/Algorithm.ts
```diff
@@ -119,6 +119,16 @@ export default class Algorithm extends Builder {
       }
     }
 
+    for (const step of node.querySelectorAll('li')) {
+      if (/\ba new Abstract Closure\b/.test(ownTextContent(step))) {
```
bakkot: There are other phrasings, e.g. "[a new Job Abstract Closure](https://tc39.es/ecma262/multipage/structured-data.html#sec-enqueueresolveinagentjob)".

We have an existing test [here](https://github.com/tc39/ecmarkup/blob/b0ebb9f005aa37f6cfd2cfc09edeae19c4b42b44/src/lint/rules/variable-use-def.ts#L249). You could factor it out and use it in both places instead of duplicating.

(There's another test [here](https://github.com/tc39/ecmarkup/blob/b0ebb9f005aa37f6cfd2cfc09edeae19c4b42b44/src/utils.ts#L289), come to think, which should probably use the "when called" test; that captures non-AC code blocks like [read-modify-write modification function](https://tc39.es/ecma262/multipage/structured-data.html#sec-atomiccompareexchangeinsharedblock) as well.)

michaelficarra:
Comments addressed.

PR #689, opened 2026-04-08
reject AO headers with duplicate parameter names

### LINKED ISSUE #687: reject AOs with duplicate parameter names
I introduced an error into my spec where I had an AO with more than one parameter with the same name, but ecmarkup didn't complain about it. The build should have failed.

michaelficarra:
Fixes #687. Disclaimer: I used AI in the process of creating this PR.

### DIFF
--- a/src/lint/rules/variable-use-def.ts
+++ b/src/lint/rules/variable-use-def.ts
@@ -144,8 +144,11 @@ export function checkVariableUsage(
       preceding.textContent != null
     ) {
       const isParameter = preceding.nodeName === 'H1';
+      const seen = new Set<string>();
       // `__` is for <del>_x_</del><ins>_y_</ins>, which has textContent `_x__y_`
       for (const name of preceding.textContent.matchAll(/(?<=\b|_)_([a-zA-Z0-9]+)_(?=\b|_)/g)) {
+        if (seen.has(name[1])) continue;
+        seen.add(name[1]);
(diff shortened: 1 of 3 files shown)


bakkot:
So I guess https://github.com/tc39/ecmarkup/pull/689 _should have_ fixed this, but instead it was erroring out for lack of location information, and it's easiest to do the check elsewhere instead of getting the error location right? Sure I guess. Kinda inelegant. Not sure you would have accepted this PR if I'd sent it. Keeping track of the error location is probably annoying I suppose.

michaelficarra:
I don't care, either one works for me.

src/lint/rules/variable-use-def.ts
```diff
@@ -144,8 +144,11 @@ export function checkVariableUsage(
       preceding.textContent != null
     ) {
       const isParameter = preceding.nodeName === 'H1';
+      const seen = new Set<string>();
       // `__` is for <del>_x_</del><ins>_y_</ins>, which has textContent `_x__y_`
       for (const name of preceding.textContent.matchAll(/(?<=\b|_)_([a-zA-Z0-9]+)_(?=\b|_)/g)) {
+        if (seen.has(name[1])) continue;
```
bakkot, line 151:
```suggestion
        // We avoid dealing with parameter re-declaration here because tracking location information is annoying. It's handled in `checkDuplicateParam` in header-parser instead.
        if (seen.has(name[1])) continue;
```

bakkot (approved):
Eh, sure,

PR #692, opened 2026-04-08
add a new lint for preferring `throw` over `return ThrowCompletion(...)`

michaelficarra:


src/lint/rules/prefer-throw-shorthand.ts
```diff
@@ -0,0 +1,49 @@
...
+  if (stepSeq == null) {
+    return;
+  }
+  const stepSource = algorithmSource.slice(
+    step.contents[0].location.start.offset,
+    step.contents[step.contents.length - 1].location.end.offset,
+  );
+  const baseOffset = step.contents[0].location.start.offset;
+
+  walkExpr(expr => {
```
bakkot, line 32: You don't need `walkExpr` for this; you only really need that when descending recursively into expressions. Here `return ThrowCompletion(...)` can only appear at the top level (i.e., not within another expression).

Instead you can just walk over `stepSeq` directly, looking for `text` nodes ending in `return ` followed by `call` `ThrowCompletion` nodes instead of reading the raw source, which is also better because parsed source accounts for `<ins>` and `<del>` and etc.

michaelficarra: Done.

bakkot (approved):

PR #693, opened 2026-04-08
add new lints matching some ECMA-262 editorial conventions

michaelficarra:


src/lint/rules/algorithm-spelling.ts
```diff
@@ -0,0 +1,133 @@
+import type { Reporter } from '../algorithm-error-reporter-type';
+import type { OrderedListItemNode } from 'ecmarkdown';
+import type { Seq } from '../../expr-parser';
+import { walk as walkExpr } from '../../expr-parser';
+import { offsetToLineAndColumn } from '../../utils';
+
+const ruleId = 'algorithm-spelling';
+
+// Note that these will be composed, so cannot contain backreferences
```
bakkot, line 9: I mean... named backreferences are fine as long as you're running on an engine with https://github.com/tc39/ecma262/pull/2721!

Probably the LLM copied this from [here](https://github.com/tc39/ecmarkup/blob/daff9b0628b397bcde12c3a002a875a401dcc1e4/src/lint/collect-spelling-diagnostics.ts#L8), but that comment predates the above spec change.

bakkot (approved):
These seem like they might have false positives, but if you've run it against 262, 402, and Temporal with no false positives it's fine to land and we can go back and relax any annoying ones later.

Not totally sure all of these should be confined to algorithms, though.

src/lint/rules/algorithm-spelling.ts
michaelficarra: Still a shared namespace. I think the comment should remain.

michaelficarra:
I ran it against 262, hence https://github.com/tc39/ecma262/pull/3805. Haven't run it against 402 or any other specs yet, but I imagine it will just discover changes we actually want to make.

src/lint/rules/algorithm-spelling.ts
bakkot: No it's not?

michaelficarra: Oh I see what you mean, you're right.

src/lint/rules/algorithm-spelling.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  // check "or if" / "and if" — but only when the other connective is not also present
+  const orIfPattern = /(?<=\bIf .+) (or|and) if\b/gu;
+  let orIfMatch = orIfPattern.exec(stepSource);
+  while (orIfMatch !== null) {
+    const otherPattern = orIfMatch[1] === 'or' ? /\band\b/ : /\bor\b/;
+    if (!otherPattern.test(stepSource)) {
+      report({
+        ruleId,
+        message: 'prefer "if _a_ or _b_" over "if _a_ or if _b_" (same for "and")',
```
nicolo-ribaudo, line 112:
```suggestion
        message: `prefer "if _a_ ${orIfMatch[1]} _b_" over "if _a_ ${orIfMatch[1]} if _b_"`,
```

nicolo-ribaudo (approved):

src/lint/rules/algorithm-spelling.ts
michaelficarra: I even thought of this when writing it, but decided not to out of laziness. Done.

PR #694, opened 2026-04-20
Make a sticky preference for single-page vs. multi-page views

### LINKED ISSUE #631: Multipage should be more accessible
Having multipage support is great, but it should be more accessible... for example, loading the single-page https://tc39.es/ecma262/ from a linked term in https://tc39.es/ecma402/ or loading a build preview such as https://ci.tc39.es/preview/tc39/ecma262/pull/3513 from a GitHub Actions CI link on a Chrome DevTools "Slow 4G" connection:
* consumes at least 8.7 MB of RAM
* takes 9 seconds to get there
* never indicates the presence of a multipage alternative
* does not respond to the "m" shortcut for switching to multipage until after all page HTML has been parsed
* responds to the "m" shortcut by navigating first to https://tc39.es/ecma262/multipage/#sec-… and then, after _that_ page has been parsed, re-navigating to https://tc39.es/ecma262/multipage/$page.html#sec-…

The "never indicates multipage" issue could be solved with a floating link, possibly cleared/moved/otherwise de-emphasized on completion of page load (cf. https://html.spec.whatwg.org/ ).

The "after HTML has been parsed" issues stem from use of `<script defer>`, which I think would be better behaved as `<script async>` or possibly even loading [`doShortcut`](https://github.com/tc39/ecmarkup/blob/44bb83e0afbe69e1784b144a8ebef4c20df0ba6b/js/menu.js#L1075) as a blocking script before the body (and probably updating it to tolerate an undefined `idToSection`).

The re-navigation issue is the smallest of these, but could be solved by loading multipage.js (specifically for `idToSection` and the [re-navigation logic](https://github.com/tc39/ecmarkup/blob/44bb83e0afbe69e1784b144a8ebef4c20df0ba6b/js/multipage.js#L14-L17)) even on the single-page view for use by `doShortcut`.

gibson042:
Fixes #631

Expose buttons in the shortcuts popup for setting a localStorage "multipagePreference" entry to either empty or "single-page" or "multi-page" (and a way for similar `data-multipage-preference="…"` controls to be embedded in documents, necessary for mobile and other keyboard-free interactions), and check for that preference when loading any page to perform a prompt redirect if needed (e.g., before loading an entire single-page spec).

Redirects are not applied when following internal links, so using toggling between single-page and multi-page views _without_ changing the preference still works.

### DIFF
--- a/js/multipage.js
+++ b/js/multipage.js
@@ -53,3 +53,27 @@ let maybeToggleMultipage = e => {
   }
 };
 document.addEventListener('keypress', maybeToggleMultipage);
+
+let maybeSetMultipagePreference = e => {
+  if (!(e.target instanceof HTMLElement)) {
+    return;
+  }
+  let target = e.target;
+  let toggler = target.closest('[data-prefer-multipage]');
+  if (!toggler || target.isContentEditable) {
+    return;
+  }
+  switch (target.nodeName.toLowerCase()) {
+    case 'textarea':
+    case 'select':
+      return;
+    case 'input': {
+      let isCheckable = target.type === 'checkbox' || target.type === 'radio';
+      if (toggler !== target || !isCheckable || !target.checked) {
+        return;
+      }
+    }
+  }
+  storage.preferMultipage = toggler.dataset.preferMultipage;
+};
+document.addEventListener('click', maybeSetMultipagePreference);
--- a/js/multipage.js
+++ b/js/multipage.js
@@ -7,13 +7,49 @@ for (let [section, ids] of Object.entries(multipageMap)) {
     }
   }
 }
-if (location.hash) {
-  let targetSec = idToSection[location.hash.substring(1)];
-  if (targetSec != null) {
-    let match = location.pathname.match(/([^/]+)\.html?$/);
-    if ((match != null && match[1] !== targetSec) || location.pathname.endsWith('/multipage/')) {
-      window.navigating = true;
-      location = (targetSec === 'index' ? './' : targetSec + '.html') + location.hash;
-    }
+let pathParts = location.pathname.split('/');
+let isMultipage = pathParts[pathParts.length - 2] === 'multipage';
+let activeSec = isMultipage ? pathParts[pathParts.length - 1].replace(/\.html$/, '') : undefined;
+let activeSecHash =
+  activeSec && idToSection['sec-' + activeSec] != null ? '#sec-' + activeSec : undefined;
+let storage = typeof localStorage !== 'undefined' ? localStorage : Object.create(null);
+
+{
+  let resolvedHash = location.hash || activeSecHash || '';
+  let targetSec = resolvedHash ? idToSection[resolvedHash.substring(1)] : undefined;
+  let preferMultipage = storage.preferMultipage;
+  if (isMultipage && preferMultipage === 'false') {
+    window.navigating = true;
+    location = pathParts.slice(0, -2).join('/') + '/' + resolvedHash;
+  } else if (
+    isMultipage
+      ? targetSec != null && (activeSec || 'index') !== targetSec
+      : preferMultipage === 'true'
+  ) {
+    window.navigating = true;
+    location = 'multipage/' + targetSec + '.html' + location.hash;
   }
 }
+let maybeToggleMultipage = e => {
+  if (!(e.target instanceof HTMLElement)) {
+    return;
+  }
+  let target = e.target;
+  let name = target.nodeName.toLowerCase();
+  if (name === 'textarea' || name === 'input' || name === 'select' || target.isContentEditable) {
+    return;
+  }
+  if (e.altKey || e.ctrlKey || e.metaKey || e.key !== 'm') {
+    return;
+  }
+  let hash = location.hash;
+  if (isMultipage) {
+    storage.preferMultipage = 'false';
+    location = pathParts.slice(0, -2).join('/') + '/' + (hash || activeSecHash || '');
+  } else {
+    storage.preferMultipage = 'true';
+    let targetSec = hash ? idToSection[hash.substring(1)] : undefined;
+    location = 'multipage/' + (targetSec ? targetSec + '.html' : '') + hash;
+  }
+};
+document.addEventListener('keypress', maybeToggleMultipage);
--- a/js/multipage.js
+++ b/js/multipage.js
@@ -100,6 +100,13 @@ function toggleMultipage() {
 
 // redirect to single-page/multi-page per preference
 (() => {
+  let { pathPrefix, isMultipage, section: activeSec } = parseSpecPath(location);
+  let activeSecHash =
+    activeSec && idToSection['sec-' + activeSec] != null ? '#sec-' + activeSec : undefined;
+  let hash = location.hash;
+  let resolvedHash = hash || activeSecHash || '';
+  let targetSec = resolvedHash ? idToSection[resolvedHash.substring(1)] : undefined;
+
   // ...except from internal links
   let referrer;
   try {
@@ -110,13 +117,6 @@ function toggleMultipage() {
   if (referrer && referrer.host === location.host) {
     if (parseSpecPath(referrer).pathPrefix === pathPrefix) return;
   }
-
-  let { pathPrefix, isMultipage, section: activeSec } = parseSpecPath(location);
-  let activeSecHash =
-    activeSec && idToSection['sec-' + activeSec] != null ? '#sec-' + activeSec : undefined;
-  let hash = location.hash;
-  let resolvedHash = hash || activeSecHash || '';
-  let targetSec = resolvedHash ? idToSection[resolvedHash.substring(1)] : undefined;
 
   let storage = window.localStorage || Object.create(null);
   let multipagePreference = getMultipagePreference(storage);
--- a/js/multipage.js
+++ b/js/multipage.js
@@ -1,4 +1,6 @@
 'use strict';
+
+// initialize globals
 let idToSection = Object.create(null);
 for (let [section, ids] of Object.entries(multipageMap)) {
   for (let id of ids) {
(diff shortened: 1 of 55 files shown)


michaelficarra:
I don't like this. I prefer using the single-page document. When I open https://tc39.es/ecma262 in a new tab, I want to stay in the single-page document. But when I click a link that somebody posts to the multipage document, I am fine with it loading the multipage document. And when I want to share a link with somebody, I sometimes switch to the multipage document to copy the URL. But I don't want to go to the multipage document the next time I load the spec. For the latter problem, I could see adding a "copy multipage document link to clipboard" keyboard shortcut, but I don't know if there's a solution to the former problem.

gibson042:
> I don't like this. I prefer using the single-page document. When I open https://tc39.es/ecma262 in a new tab, I want to stay in the single-page document. But when I click a link that somebody posts to the multipage document, I am fine with it loading the multipage document. And when I want to share a link with somebody, I sometimes switch to the multipage document to copy the URL. But I don't want to go to the multipage document the next time I load the spec. For the latter problem, I could see adding a "copy multipage document link to clipboard" keyboard shortcut, but I don't know if there's a solution to the former problem.

Could you further elaborate upon "the former problem"? When following a link to the multi-page document, are you **_fine_** with loading the multi-page resource, or are you **_insisting_** upon that? And for links to the single-page document, would you likewise insist upon loading _that_ rather than redirecting per preference to the multi-page document? Because if not, then it seems like either the first problem collapses into the second[^1] or that you want the preference to asymmetrically affect only requests for the single-page document[^2], but if you **would** insist upon loading the single-page document then it seems like you reject at least part of #631 and I'd like to explore that.

[^1]: solved by e.g. single-page vs. multi-page permalink access and/or a more explicit way to set the preference
[^2]: which is certainly possible, but less friendly to those with a single-page preference

bakkot:
I would prefer links to the single-page document always open as single-page and links to the multi-page document always open as multi-page. Also, like @michaelficarra I regularly toggle between the two for the purposes of sharing links or find-on-page, without that indicating a preference about how future pages should be opened. I'm OK with having some sort of setting for "would you like to override opened links to force one state or the other", but I don't think it is appropriate for merely switching between the two states to be taken to indicate such a preference.

ljharb:
I only ever want the single-page version, but I agree with https://github.com/tc39/ecmarkup/pull/694#issuecomment-4281993953

gibson042:
> I would prefer links to the single-page document always open as single-page and links to the multi-page document always open as multi-page. Also, like @michaelficarra I regularly toggle between the two for the purposes of sharing links or find-on-page, without that indicating a preference about how future pages should be opened. I'm OK with having some sort of setting for "would you like to override opened links to force one state or the other", but I don't think it is appropriate for merely switching between the two states to be taken to indicate such a preference.

Thanks, that's useful feedback. How would y'all feel about the preference being set from [About this Specification](https://tc39.es/ecma262/multipage/#:~:text=This%20document%20is%20available%20as%20a%20single%20page%20and%20as%20multiple%20pages.), such that the experience of updating it is something like "_scroll/navigate to the start of the spec and then click as appropriate to set the preference and reload_"? Maybe buttons that are enabled in correspondence with `localStorage`?

> This document is available as <a href="https://tc39.es/ecma262/">a single page</a> and as <a href="https://tc39.es/ecma262/multipage/">multiple pages</a>. You can set a permanent preference for which view to use: <ins>**No preference**</ins> or <ins>Always single-page</ins> or <ins>Always multi-page</ins>.

It could also be added to the keyboard shortcuts popup for convenience, but that's too difficult to access on mobile so it couldn't live _only_ there... and I'm inclined to leave it out.

bakkot:
WFM. I'd also put it in the shortcuts menu personally but don't feel strongly.

gibson042:
I have made the change to set the preference only by direct interaction with elements explicitly marked for that purpose.

michaelficarra:
What happens when you press `m` while one of these options is selected? Does it navigate you away and right back? Does it do nothing? Does it change your preference?

I'm still not convinced by the utility of this feature. If it stays completely out of the way of my workflow and the workflows of consumers who don't use it, I don't have much standing to object to it, but at this point I still don't see the point of it.

gibson042:
> What happens when you press `m` while one of these options is selected? Does it navigate you away and right back? Does it do nothing? Does it change your preference?

As requested above, there is no change in what happens when you press `m`—it toggles the view without touching the preferences, allowing you to freely bounce between single-page and multi-page with no effect on future visits.

> I'm still not convinced by the utility of this feature. If it stays completely out of the way of my workflow and the workflows of consumers who don't use it

It does; anyone who leaves the preference unset will have no change in experience.

> I don't have much standing to object to it, but at this point I still don't see the point of it.

The point is to improve the experience for anyone who does have a preference, particularly on mobile devices which often have limited resources and are effectively precluded from using keyboard shortcuts.

michaelficarra:
> As requested above, there is no change in what happens when you press `m`—it toggles the view without touching the preferences, allowing you to freely bounce between single-page and multi-page with no effect on future visits.

Doesn't that kind of violate the preference?

> The point is to improve the experience for anyone who does have a preference, particularly on mobile devices which often have limited resources and are effectively precluded from using keyboard shortcuts.

Since the redirect happens client-side, following a link to the single-page document will still load the single-page document before the multi-page one, actually increasing data use. Perhaps setting a cookie so the server could 302/307 to the appropriate page would be a better implementation if that's the goal.

ljharb:
the server is github pages, there's no logic that can run there.

michaelficarra:
I'm just talking about ideals.

gibson042:
> > As requested above, there is no change in what happens when you press `m`—it toggles the view without touching the preferences, allowing you to freely bounce between single-page and multi-page with no effect on future visits.
> 
> Doesn't that kind of violate the preference?

No, we're treating `m` as an explicit signal to interact in a specific mode regardless of preference (https://github.com/tc39/ecmarkup/pull/694#issuecomment-4281993953 : "_I regularly toggle between the two for the purposes of sharing links or find-on-page, **without that indicating a preference about how future pages should be opened**_").

> > The point is to improve the experience for anyone who does have a preference, particularly on mobile devices which often have limited resources and are effectively precluded from using keyboard shortcuts.
> 
> Since the redirect happens client-side, following a link to the single-page document will still load the single-page document before the multi-page one, actually increasing data use.

It will load _part_ of the single-page document, and then abort to load the multi-page document ASAP (because the first requested external resource is multipage.js via a `<script>` element with attribute `async`). For ECMA-262, overall data use will be lower than loading the single-page document (and likewise with respect to consumption of other resources such as memory and time).

> Perhaps setting a cookie so the server could 302/307 to the appropriate page would be a better implementation if that's the goal.

ecmarkup produces content to be served from many places, and cannot rely upon server-side support (which in practice is effectively limited to static content anyway—neither ECMA nor GitHub are likely to open up cookie-based server-side redirection). And to be clear, the goal is practical improvement for ECMA-262 without being so narrow that it doesn't also apply to other documents availing theirselves of multipage support.

js/multipage.js
```diff
@@ -7,13 +17,91 @@ for (let [section, ids] of Object.entries(multipageMap)) {
...
+    for (let el of document.querySelectorAll('[disabled][data-multipage-preference]')) {
+      el.disabled = false;
     }
+  };
+  if (document.readyState === 'loading') {
+    document.addEventListener('DOMContentLoaded', enableToggles);
+  } else {
+    enableToggles();
   }
+  document.addEventListener('click', e => {
```
nicolo-ribaudo: Can we just listen on the `fieldset` that contains the three `[data-multipage-preference]` elements, or on the elements themselves?

michaelficarra:
Ask from editor call was

- [x] make the click handler less generic, only handling the cases that ecmarkup-generated UI is responsible for
- [x] expose global functions for interacting with this preference so that specs can add their own custom UIs
- [x] split out the formatting run into a separate PR

gibson042:
> Ask from editor call was
> 
> - [x] make the click handler less generic, only handling the cases that ecmarkup-generated UI is responsible for
> - [x] expose global functions for interacting with this preference so that specs can add their own custom UIs
> - [x] split out the formatting run into a separate PR

Done, along with making the multipage preference per-document (similar to pins as of #705, but defaulting PR-preview documents to their ancestor spec).

js/multipage.js
```diff
@@ -7,13 +19,132 @@ for (let [section, ids] of Object.entries(multipageMap)) {
...
+    window.navigating = true;
+    location = 'multipage/' + (targetSec ? targetSec + '.html' : '') + location.hash;
+  }
+})();
+
+// enable preference togglers
+document.documentElement.dataset.multipagePreference = getMultipagePreference(window.localStorage);
+if (window.localStorage) {
+  let storage = window.localStorage;
+  document.addEventListener('click', e => {
```
michaelficarra: Why are we still putting the event listener on `document` rather than on the buttons themselves (or at least their parent)?

ljharb: using delegation on a root element is much more performant than per-element listeners

michaelficarra: Okay but why not their nearest shared parent instead of all the way at the document?

ljharb: fair point

gibson042: > Why are we still putting the event listener on `document` rather than on the buttons themselves (or at least their parent)?

multipage.js is requested in the \<head> and is generally expected to load before <tt>#shortcuts-help</tt> exists (but is not guaranteed to do so).

Would you prefer another approach, e.g. `buildShortcutsHelp` emitting HTML with an inline \<script>, or perhaps this code dynamically choosing between adding click listeners immediately vs. in a DOMContentLoaded callback?

michaelficarra: IME it's normal to just add event listeners in a DOMContentLoaded listener (or move the script to the end when it's inline)

gibson042: It's important for the rest of this script to be as early as possible, although this event listener wiring could be moved into menu.js (which is entirely deferred to follow DOMContentLoaded). However, I'd rather not separate definition and use of `setMultipagePreference` like that.

michaelficarra: I don't understand. What I'm suggesting here is to replace this addEventListener call with a DOMContentLoaded listener that, in its handler, adds a click listener to `#shortcuts-help`. You wouldn't need to move anything.

gibson042: I've pushed the DOMContentLoaded approach; please comment with your preferred alternative if that still doesn't suit you.

js/multipage.js
```diff
@@ -7,13 +19,150 @@ for (let [section, ids] of Object.entries(multipageMap)) {
...
+  };
+
+  let shortcuts = document.getElementById('shortcuts-help');
+  if (shortcuts) {
+    enableToggles(shortcuts);
+  } else {
+    document.addEventListener('DOMContentLoaded', () => {
+      shortcuts = document.getElementById('shortcuts-help');
+      if (shortcuts) enableToggles(shortcuts);
+    });
```
michaelficarra:
```suggestion
  let enableToggles = container => {
    container.addEventListener('click', e => {
      let target = e.target;
      if (!target.matches?.('button[data-multipage-preference]')) {
        return;
      }
      e.preventDefault();
      let preference = e.currentTarget.getAttribute('data-multipage-preference');
      if (typeof preference !== 'string') {
        return;
      }
      document.documentElement.setAttribute('data-multipage-preference', preference);
      setMultipagePreference(storage, preference);
    });
  };

  let shortcuts = document.getElementById('shortcuts-help');
  if (shortcuts) {
    enableToggles(shortcuts);
  } else {
    document.addEventListener('DOMContentLoaded', () => {
      shortcuts = document.getElementById('shortcuts-help');
      if (shortcuts) enableToggles(shortcuts);
    });
  }
```

but I would do it all in the DOMContentLoaded listener personally

```suggestion
  document.addEventListener('DOMContentLoaded', () => {
    let shortcuts = document.getElementById('shortcuts-help');
    if (shortcuts == null) {
      return;
    }
    shortcuts.addEventListener('click', e => {
      let target = e.target;
      if (!target.matches?.('button[data-multipage-preference]')) {
        return;
      }
      e.preventDefault();
      let preference = e.currentTarget.getAttribute('data-multipage-preference');
      if (typeof preference !== 'string') {
        return;
      }
      document.documentElement.setAttribute('data-multipage-preference', preference);
      setMultipagePreference(storage, preference);
    });
  });
```

gibson042: I just pushed a version of the first suggestion (patched to resolve bugs that I observed in local testing).

js/multipage.js
```diff
@@ -7,13 +19,156 @@ for (let [section, ids] of Object.entries(multipageMap)) {
...
+      let target = e.target.closest?.(`[${SET_MULTIPAGE_PREFERENCE_ATTR}]`);
+      let preference = target?.getAttribute(SET_MULTIPAGE_PREFERENCE_ATTR);
+      if (typeof preference !== 'string') {
+        return;
+      }
+      document.documentElement.setAttribute(MULTIPAGE_PREFERENCE_ATTR, preference);
+      setMultipagePreference(storage, preference);
+    });
+    // Work around a bug where contents are sometimes empty.
+    setTimeout(() => {
```
michaelficarra, line 158: This is super sketchy but okay.

michaelficarra (approved):

js/multipage.js
gibson042: Absolutely agreed; I was _shocked_ to see an observably empty shortcuts-help div.

PR #695, opened 2026-05-03
Improve searchbox event handling

gibson042:
* use "input" rather than "keyup" to respond to value changes, catching previously ignored interactions such as Paste and avoiding the need to special-case Enter/Tab/etc. since those don't trigger an "input" event
* call `preventDefault` in the synchronous listener rather than in the debounced function when it's too late
* drop attempted "/" suppression, since it wasn't working anyway and apparently hasn't been a problem
* consolidate "keydown"/"input" handling into a single debounced listener that switches on event type for deciding its behavior

### DIFF
--- a/js/menu.js
+++ b/js/menu.js
@@ -9,14 +9,17 @@ function Search(menu) {
 
   document.addEventListener('keydown', this.documentKeydown.bind(this));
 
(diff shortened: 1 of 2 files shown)


michaelficarra (approved):
Thanks!

PR #699, opened 2026-06-01
Update styles to cycle through decimal/alpha/roman until nesting level 9

linusg:
I'm told the current limit of using roman markers for any nesting level above 6 is from the MS Word days. Current day MS Word and Google Docs limit the nesting level of lists to 9 and fully cycle through decimal/alpha/roman until then.

ECMA-262's maximum nesting level is 8 at the moment (GDI, EDI), leading to repeated roman markers. Short of refactoring that out of existence it seems reasonable to add proper support for nesting levels up to 9 and use a generic fallback beyond that. If you ever end up there something went very wrong.

While this will break any reference like "GlobalDeclarationInstantiation step 12.b.ii.2.a.ii.ii.ii", those references have always been brittle unless a specific snapshot is referenced. Could be step 15.c.... next week without this change too.

Before (leading 12. cut off):

image

After:

image

### DIFF
--- a/css/elements.css
+++ b/css/elements.css
@@ -487,7 +487,14 @@ emu-alg ol,
 emu-alg ol ol ol ol,
 emu-alg ol.nested-thrice,
 emu-alg ol.nested-twice ol,
-emu-alg ol.nested-once ol ol {
+emu-alg ol.nested-once ol ol,
+/* depth 7 */
+emu-alg ol ol ol ol ol ol ol,
+emu-alg ol.nested-lots ol,
+emu-alg ol.nested-four-times ol ol,
+emu-alg ol.nested-thrice ol ol ol,
+emu-alg ol.nested-twice ol ol ol ol,
+emu-alg ol.nested-once ol ol ol ol ol {
   list-style-type: decimal;
 }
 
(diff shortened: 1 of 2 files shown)


css/elements.css
```diff
@@ -514,11 +528,26 @@ emu-alg ol.nested-four-times ol,
...
 
+/* depth 10+ fallback */
+emu-alg ol ol ol ol ol ol ol ol ol ol,
+emu-alg ol.nested-lots ol ol ol ol ol,
+emu-alg ol.nested-four-times ol ol ol ol ol ol,
+emu-alg ol.nested-thrice ol ol ol ol ol ol ol,
+emu-alg ol.nested-twice ol ol ol ol ol ol ol ol,
+emu-alg ol.nested-once ol ol ol ol ol ol ol ol ol {
+  list-style-type: disc;
+}
```
gibson042: I don't think we should ever present an ordered list as unordered; might as well keep the roman-numeral fallback.
```suggestion
/* depth 10+ fallback */
emu-alg ol ol ol ol ol ol ol ol ol ol,
emu-alg ol.nested-lots ol ol ol ol ol,
emu-alg ol.nested-four-times ol ol ol ol ol ol,
emu-alg ol.nested-thrice ol ol ol ol ol ol ol,
emu-alg ol.nested-twice ol ol ol ol ol ol ol ol,
emu-alg ol.nested-once ol ol ol ol ol ol ol ol ol {
  list-style-type: lower-roman;
}
```

gibson042 (approved):

css/elements.css
linusg: Fixed by folding the fallback into the `depth 9` rules.

michaelficarra (approved):

rbuckton:
Have we considered using a `@counter-style` rule instead? Something like:

```css
@counter-style algorithm-steps {
  system: cyclic;
  symbols: decimal lower-alpha lower-roman;
}

emu-alg ol {
  list-style-type: algorithm-steps;
}
```

michaelficarra:
@rbuckton We could investigate that with a follow-up PR, but I suspect we'll have to keep most of the current CSS anyway to support the nested-n-times classes.

PR #701, opened 2026-06-16
When formatting, replace inline ASCII quotes with typographic quotes

gibson042:
Ref https://github.com/tc39/ecma262/pull/3861#pullrequestreview-4412001453
Ref #173
Ref #317

Note that this processing cannot be scoped to individual text nodes because logical quotations can span across those, e.g. <tt><ins>a "</ins>\<a href="…"><ins>binary64</ins>\</a><ins> value"</ins></tt> should get rewritten into <tt><ins>a “</ins>\<a href="…"><ins>binary64</ins>\</a><ins> value”</ins></tt> (but the same would not be true if the medial element were block-level rather than inline).

ASCII quotes are not replaced inside of HTML comments, `<code>`/`<emu-val>` elements, backtick spans (e.g., ``` `code` ```), asterisk spans (e.g., `*"string"*`), or after equals signs (as in HTML element attributes).

This approach is not perfect, but spec source text tends to strongly avoid the sort of edge cases that would reveal its flaws (_e.g., using it as an automated analog of https://github.com/tc39/ecma262/pull/3861 finds only one line to change, in [substring](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#substring)—from «If the <del>"to"</del> suffix is omitted» to «If the **“to”** suffix is omitted»_).

### DIFF
--- a/test/formatter.ts
+++ b/test/formatter.ts
@@ -261,16 +261,18 @@ describe('document formatting', () => {
     );
   });
 
-  it('<script>, <style>, and <pre> whitespace is preserved', async () => {
+  it('whitespace and inner ASCII quotes are preserved inside <script>, <style>, and <pre>', async () => {
     await assertRoundTrips(
       `<script>
     content
+    "quoted content"
 </script>
 <style>
     content
+    "quoted content"
 </style>
 <pre>
-    content
+    "content"
 </pre>
 `,
     );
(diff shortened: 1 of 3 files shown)


michaelficarra:
In editor call today, I asked to confirm that this auto-formatting also would have made all of the changes from https://github.com/tc39/ecma262/pull/3861 for us automatically.

michaelficarra:
Also, do we have tests for the formatter that we can add to? In particular, I think the negative cases will be valuable.

bakkot:
Yes: https://github.com/tc39/ecmarkup/blob/39c401c8ba885942c142cfff513a9d5476d83640/test/formatter.ts

(Including negative tests in the form of `assertRoundTrips`.)

michaelficarra:
Then let's also add tests for this @gibson042.

gibson042:
> In editor call today, I asked to confirm that this auto-formatting also would have made all of the changes from [tc39/ecma262#3861](https://github.com/tc39/ecma262/pull/3861) for us automatically.

This was a great idea, and did in fact find some bugs (now fixed).

> Then let's also add tests for this @gibson042.

Done.

test/formatter.ts
```diff
@@ -324,6 +326,39 @@ describe('document formatting', () => {
...
+      <emu-alg>
+        1. [x="a"] Assert: Quotes are also “detected” in “algorithm steps”.
+      </emu-alg>
+      `,
+    );
+  });
+
+  it('preserves ASCII quotes that are code', async () => {
+    await assertRoundTrips(
+      `<p>ASCII quotes are not replaced in <code>"code elements"</code>, <emu-val>"emu-val elements"</emu-val>, ${'`'}"backtick spans"${'`'}, or *"inline language strings"*.</p>\n`,
```
michaelficarra: Also put a quoted HTML attribute in this test.

test/formatter.ts
```diff
@@ -261,16 +261,18 @@ describe('document formatting', () => {
     );
   });
 
-  it('<script>, <style>, and <pre> whitespace is preserved', async () => {
+  it('<script>, <style>, and <pre> whitespace and inner ASCII quotes are preserved', async () => {
```
michaelficarra:
```suggestion
  it('<script>, <style>, <pre> whitespace, and inner ASCII quotes are preserved', async () => {
```

nicolo-ribaudo (approved):

michaelficarra (approved):

PR #705, opened 2026-06-26
isolate pin storage by path

### LINKED ISSUE #702: pins are not isolated by path
This means that pins are shared between 262/402/etc and also between the draft spec and PR previews (everything on tc39.es). The observed behaviour is that every time you load a document on this domain, it clears any pins for sections that that document doesn't have. Which is often the case for PRs that add sections (you will often pin the added section) or when switching between 262 and one of the other specs (all pins are cleared).

michaelficarra:
Fixes #702. Had to also do some extra bookkeeping so we don't grow localStorage indefinitely.

### DIFF
--- a/js/menu.js
+++ b/js/menu.js
@@ -486,7 +486,6 @@ Menu.prototype.addPinEntry = function (id) {
   if (!entry) {
     // id was deleted after pin (or something) so remove it
     delete this._pinnedIds[id];
-    this.persistPinEntries();
     return;
   }
 
@@ -510,7 +509,6 @@ Menu.prototype.addPinEntry = function (id) {
     this.showPins();
   }
   this._pinnedIds[id] = true;
-  this.persistPinEntries();
 };
 
 Menu.prototype.removePinEntry = function (id) {
@@ -520,14 +518,13 @@ Menu.prototype.removePinEntry = function (id) {
   if (Object.keys(this._pinnedIds).length === 0) {
     this.hidePins();
   }
-
-  this.persistPinEntries();
 };
 
 Menu.prototype.unpinAll = function () {
   for (let id of Object.keys(this._pinnedIds)) {
     this.removePinEntry(id);
   }
+  this.persistPinEntries();
 };
 
 Menu.prototype.pinListClick = function (event) {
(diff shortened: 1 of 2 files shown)


js/menu.js
```diff
@@ -546,7 +605,16 @@ Menu.prototype.persistPinEntries = function () {
     return;
   }
 
-  localStorage.pinEntries = JSON.stringify(Object.keys(this._pinnedIds));
+  let store = this.prunePinStore(this.readPinStore());
```
gibson042: Suggestion: replace the nullary `readPinStore` with a unary `parsePinEntries` that is ignorant of `localStorage`.
```suggestion
  let store = this.parsePinEntries(window.localStorage[PIN_STORAGE_KEY]);
  this.prunePinStore(store);
```

js/menu.js
```diff
@@ -556,12 +624,20 @@ Menu.prototype.loadPinEntries = function () {
...
+  this.writePinStore(store);
+  if (!entry) return;
+
+  // addPinEntry only updates in-memory state + the DOM (and drops ids missing from
+  // this document); commit once afterwards to bump lastUsed and store the cleaned
+  // set of ids.
+  for (let i = 0; i < entry.pins.length; i++) {
+    this.addPinEntry(entry.pins[i]);
   }
+  this.persistPinEntries();
```
gibson042:
```suggestion
  let store = this.parsePinEntries(window.localStorage[PIN_STORAGE_KEY]);
  let entry = store[this.pinDocumentPath()] || { pins: [] };

  // Update in-memory state (including dropping missing ids) and the DOM.
  for (let i = 0; i < entry.pins.length; i++) {
    this.addPinEntry(entry.pins[i]);
  }

  // Bump lastUsed and commit any schema migration that was performed.
  this.persistPinEntries();
```

js/menu.js
```diff
@@ -520,23 +518,84 @@ Menu.prototype.removePinEntry = function (id) {
...
+  } catch (e) {
+    return {};
+  }
+  if (Array.isArray(parsed)) {
+    // Legacy format: one global array of ids shared across all documents. Ids not
+    // in this document are dropped later by addPinEntry().
+    return { [this.pinDocumentPath()]: { pins: parsed, lastUsed: Date.now() } };
+  }
+  return parsed && typeof parsed === 'object' ? parsed : {};
+};
```
gibson042: In addition the the `parsePinEntries` change, I think migration should be more accommodating of existing pins from multiple possible documents.
```suggestion
// Parse stored data into { path: { pins, lastUsed } }, migrating a legacy
// global array into likely document associations.
Menu.prototype.parsePinEntries = function (raw) {
  if (!raw) return {};
  let parsed;
  try {
    parsed = JSON.parse(raw);
  } catch (e) {
    return {};
  }
  if (Array.isArray(parsed)) {
    // Legacy format: one global array of ids shared across all documents.
    // Attribute them to all likely documents (including the current one),
    // trusting addPinEntry and prunePinStore to clean up any mistakes.
    let lastUsed = Date.now();
    let migrated = {};
    for (let spec of ['ecma262', 'ecma402', 'ecma404', 'ecma426']) {
      let key = 'https://tc39.es/' + spec + '/';
      migrated[key] = { pins: parsed, lastUsed };
    }
    migrated[this.pinDocumentPath()] = { pins: parsed, lastUsed };
    return migrated;
  }
  return parsed && typeof parsed === 'object' ? parsed : {};
};
```

js/menu.js
```diff
@@ -520,23 +518,84 @@ Menu.prototype.removePinEntry = function (id) {
...
+const PIN_STORAGE_KEY = 'pinEntries';
+// Forget a document's pins if it hasn't been visited in this long, so that the
+// unique-per-PR paths used by preview deployments don't grow localStorage without
+// bound.
+const PIN_TTL_MS = 180 * 24 * 60 * 60 * 1000; // 180 days
+
+// Identify the current document by path so pins in one spec/preview never clobber
+// another served from the same origin. Returns the path only; the localStorage key
+// is the constant above.
+Menu.prototype.pinDocumentPath = function () {
```
gibson042: `pin` can be read as a verb, and "document path" is potentially ambiguous with the `path` component of a URL. Suggestion: rename the concept.
```suggestion
// All per-document pins live in one storage entry, as an object of
// { [documentKey]: { pins: string[], lastUsed: <ms epoch> } } (see #702).
const PIN_STORAGE_KEY = 'pinEntries';
// Forget a document's pins if it hasn't been visited in this long, so that the
// unique-per-PR paths used by preview deployments don't grow localStorage without
// bound.
const PIN_TTL_MS = 180 * 24 * 60 * 60 * 1000; // 180 days

// Return the key associated with the current document in the object persisted at
// PIN_STORAGE_KEY (used to prevent pins in one spec/preview from clobbering those in
// other documents served from the same origin).
Menu.prototype.getDocumentKey = function () {
```

js/menu.js
```diff
@@ -520,23 +518,84 @@ Menu.prototype.removePinEntry = function (id) {
...
   }
+  return store;
+};
+
+Menu.prototype.writePinStore = function (store) {
+  try {
+    window.localStorage[PIN_STORAGE_KEY] = JSON.stringify(store);
+  } catch (e) {
+    // localStorage may be full or unavailable; pins are best-effort, so ignore.
+  }
```
gibson042: With the `loadPinEntries` suggestion, this can be inlined back into `persistPinEntries`.

gibson042 (approved):
I think this can be cleaned up a bit, but the fundamental approach works for me.

PR #709, opened 2026-07-09
add an exception to the for-each-of lint rule for interval iteration

### LINKED ISSUE #708: Linting should not complain about "For each $x in the inclusive interval from $a to $b"
As seen in https://github.com/tc39/ecma262/pull/3902 about https://tc39.es/ecma262/#step-repeatmatcher-clear-captures
> 4. For each integer _k_ in the inclusive interval from _parenIndex_ + 1 to _parenIndex_ + _parenCount_, set _capability_[_k_] to **undefined**.

```console
spec.html:38272:78: expected "of" instead of "in" in "for each"
```

michaelficarra:
Fixes #708

### DIFF
--- a/src/lint/rules/for-each-of.ts
+++ b/src/lint/rules/for-each-of.ts
@@ -26,22 +26,24 @@ export default function (
   // Find the loop variable (underscore), then check the text after it
   for (let i = 1; i < stepSeq.items.length - 1; i++) {
     const item = stepSeq.items[i];
-    if (item.name === 'underscore') {
-      const next = stepSeq.items[i + 1];
-      if (
-        next.name === 'text' &&
-        /^ in\b/.test(next.contents) &&
-        // Iterating over an interval is a valid construction where "in" is correct, in either the
-        // "inclusive interval from _a_ to _b_" shorthand or the full "interval from _a_ ... to _b_ ..." form.
-        !/^ in the (?:inclusive )?interval from\b/.test(next.contents)
-      ) {
-        report({
-          ruleId,
-          ...offsetToLineAndColumn(algorithmSource, next.location.start.offset + 1),
-          message: 'expected "of" instead of "in" in "for each"',
-        });
-      }
-      break;
+    if (item.name !== 'underscore') {
+      continue;
     }
+
+    const next = stepSeq.items[i + 1];
+    if (
+      next.name === 'text' &&
+      /^ in\b/.test(next.contents) &&
+      // Iterating over an interval is a valid construction where "in" is correct, in either the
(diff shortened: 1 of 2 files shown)


src/lint/rules/for-each-of.ts
```diff
@@ -27,7 +28,13 @@ export default function (
     const item = stepSeq.items[i];
     if (item.name === 'underscore') {
       const next = stepSeq.items[i + 1];
-      if (next.name === 'text' && /^ in\b/.test(next.contents)) {
+      if (
+        next.name === 'text' &&
+        /^ in\b/.test(next.contents) &&
+        // Iterating over an interval is a valid construction where "in" is correct, in either the
+        // "inclusive interval from _a_ to _b_" shorthand or the full "interval from _a_ ... to _b_ ..." form.
+        !/^ in the (?:inclusive )?interval from\b/.test(next.contents)
```
bakkot:
```suggestion
        !/^ in the (?:inclusive )?interval\b/.test(next.contents)
```
No need to mandate the "from"; e.g. one might say "in the interval defined by..." or whatever.

michaelficarra: All of our allowed interval forms use `from`: https://tc39.es/ecma262/#interval

gibson042: I second @bakkot's suggestion to stop after "interval". And if we consider it in scope, also adopting the early-exit pattern from above this loop:
```ts
  // Find the loop variable (underscore), then check the text after it
  for (let i = 1; i < stepSeq.items.length - 1; i++) {
    const item = stepSeq.items[i];
    if (item.name !== 'underscore') {
      continue;
    }

    const next = stepSeq.items[i + 1];
    const over = next.name === 'text'
      ? next.contents.match(/^ in (the (?:inclusive)? interval )?/)
      : undefined;
    if (over && !over[1]) {
      report({
        ruleId,
        ...offsetToLineAndColumn(algorithmSource, next.location.start.offset + 1),
        message: 'expected "of" instead of "in" in "for each"',
      });
    }
    break;
  }
```

gibson042 (approved):
LGTM even without accepting the suggestion(s).

src/lint/rules/for-each-of.ts
bakkot: > All of our allowed interval forms use `from`: https://tc39.es/ecma262/#interval

This rule is not intended to enforce a specific style for declaring intervals. You could have such a rule if you really wanted, but I would still want _this_ rule not to enforce it; writing "in the interval defined by..." shouldn't lead to an error complaining about using `in` instead of `of` even if there's some other error that complains about some other thing.

PR #711, opened 2026-07-14
Internal method linking

### LINKED ISSUE #116: Improve navigation of internal methods
Very often when I navigate the spec I end up grepping/searching for internal methods in order to figure out who implements a special case.
It would be nice if you could either show a quick popup with all the implementations or directly search for it in the TOC.

bakkot:
Followup to https://github.com/tc39/ecmarkup/pull/666. Fixes https://github.com/tc39/ecmarkup/issues/116.

Corresponding ecma262 PR at https://github.com/tc39/ecma262/pull/3912.

This is a breaking change because it now expects `type="internal method"` clauses, which are the clauses that define the actual implementations of the various methods on the various kinds of object, to have a corresponding definition in the tables of essential internal methods.

Claude wrote this from the following prompt, wherein I basically told it to copy https://github.com/tc39/ecmarkup/pull/666.

<details>
<summary>prompt</summary>

This is ecmarkup, the tool used to build the JS spec.

In commit 2720d8e, we added support for linking abstract/concrete methods within the spec (these have nothing to do with JS methods). These are things like `HasBinding` on Environment Records, which has multiple implementations (for a block scope vs a `with` scope, e.g.). Read that commit for context.

We want to do something similar for so-called "internal methods". These currently look like

```
    <emu-clause id="sec-ordinary-object-internal-methods-and-internal-slots-getprototypeof" type="internal method">
      <h1>[[GetPrototypeOf]] ( ): a normal completion containing either an Object or *null*</h1>
      <dl class="header">
        <dt>for</dt>
        <dd>an ordinary object _obj_</dd>
      </dl>
      <emu-alg>
        1. Return OrdinaryGetPrototypeOf(_obj_).
      </emu-alg>
```

Note that they already have a type and structured header, so there is less to do than there was for concrete methods (which did not). Also note that these always have the `[[` brackets; they are invoked as e.g. `_foo_.[[GetPrototypeOf]]()`. Otherwise, we want to take a very similar approach as in the above commit, including the new element for generating the list of implementations.
</details>

plus a couple minor followups.


src/InternalMethodDfns.ts
```diff
@@ -0,0 +1,63 @@
+import type Spec from './Spec';
```
nicolo-ribaudo, line 1: This file is almost line-by-line identical to `ConcreteMethodDfns.ts`, with the only differences being:
- the class name
- the `static readonly elements`
- `spec._internalMethodDfnsLists` vs `spec._concreteMethodDfnsLists`
- `spec.biblio.byInternalMethodAoid` vs `spec.biblio.byAbstractMethodAoid`

It'd be nice if we had a shared base class that then we use as
```js
export default class InternalMethodDfns extends MethodsBuilder {
  static readonly elements = ['EMU-INTERNAL-METHOD-DFNS'] as const;

  static _getMethodDfnsLists(spec) {
    return spec._internalMethodDfnsLists;
  }

  getDefinitions(spec, _for, namespace) {
    return spec.biblio.byInternalMethodAoid(_for, namespace);
  }
}
```

So that if we ever need to update those classes they don't risk accidentally getting out of sync.

src/Biblio.ts
```diff
@@ -38,6 +39,7 @@ class EnvRec {
     this._byProductionName = {};
     this._byAoid = {};
     this._byAbstractMethodAoid = {};
+    this._byInternalMethodAoid = {};
```
nicolo-ribaudo: Given how similar the code for the two things is, I wonder if it might make sense to have a single `this._byMethodAoid` map (they have non-conlficting aoids anyway) so that we can avoid to write the same code for the two in a bunch of places.

src/InternalMethodDfns.ts
bakkot: Yeah, I was going back and forth on this, but I guess a base class is reasonable. Will do.

bakkot: Turns out with your latter suggestion they can just be the same class.

src/Biblio.ts
bakkot: Done.

src/typechecker.ts
```diff
@@ -35,42 +35,49 @@ const getExpressionVisitor =
...
       if (biblioEntry.kind === 'syntax-directed operation') {
         warn(
           callee[0].location.start.offset,
           `${calleeName} is a syntax-directed operation and should not be invoked like a regular call`,
         );
-      } else if (biblioEntry.kind === 'abstract method' && !callIsMethod) {
+      } else if (isMethodKind && !callIsMethod) {
         warn(
           callee[0].location.start.offset,
           `${calleeName} is a method but here it is missing a record to call it on`,
```
nicolo-ribaudo: This should say "record" for abstract methods and "object" or "value" for internal methods.

nicolo-ribaudo (approved):
One minor suggestion but otherwise it looks good 👍

Issue #31, opened 2015-08-08
Include file-issue.js equivalent


domenic:
https://resources.whatwg.org/file-issue.js

In action: https://url.spec.whatwg.org/ or other specs

Would be great to be able to report issues easily in Ecmarkup specs.

bterlson:
What does the js file do?

domenic:
It inserts the link you see in the bottom right corner, and makes it so that clicking on that link with any text highlighted will quote the text in the body of the new GitHub issue. Try it out!

bterlson:
Ahh interesting. I missed that when I looked. I was actually planning something similar in addition to an edit link which takes you to the file on GitHub.

PR #719, opened 2026-08-13
Support non-ASCII characters in parameter/alias names

### LINKED ISSUE #718: Parameter/alias names are limited to ASCII
In discussion of https://github.com/tc39/ecma262/pull/3759 , @ptomato pointed out that ecmarkup restricts parameter/alias names to ASCII alphanumerics and therefore does not allow e.g. `_µs_`. Whether or not ECMA-262 is ultimately comfortable with that, it seems like ecmarkup should allow it.

gibson042:
Fixes #718

`_` is not allowed in parameter/alias names because it delimits them, but this otherwise aligns with [UAX #31](https://www.unicode.org/reports/tr31/).

The initial state does not yet include test coverage; I wanted to get this up for discussion first.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -14,7 +14,7 @@
   },
   "parserOptions": {
     "sourceType": "module",
-    "ecmaVersion": 2020
+    "ecmaVersion": 2024
   },
   "overrides": [
     {
--- a/src/Clause.ts
+++ b/src/Clause.ts
@@ -668,18 +668,18 @@ export function parseParams(params: string): ParsedParams | null {
 
   // Patterns with ..., &hellip;, or …
   if (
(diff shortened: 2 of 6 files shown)


linusg (approved):
in favor of this change, not in favor of using it in 262. :)

michaelficarra:
I think we should exclude `_`.

gibson042:
> I think we should exclude `_`.

Agreed, and that's what I have here.

michaelficarra:
Ooh, I misread where the `*` was and thought it was just prohibiting separating start/continue by underscore. Can we use set notation instead? `[\p{ID_Continue}--_]*` is more readable to me than having to count parens.

gibson042:
Sure; done.

PR #721, opened 2026-08-18
Add support for an "undeclared" step annotation

### LINKED ISSUE #720: Proposals need the ability to undeclare aliases
Noticed at https://tc39.es/proposal-intl-keep-trailing-zeros/#sec-resolveplural , in which _n_ is converted from a parameter to an internal alias that is incorrectly identified as already declared.
```
    <emu-clause id="sec-resolveplural" type="abstract operation" number="2">
      <h1>
        ResolvePlural (
          _pluralRules_: an Intl.PluralRules,
          <del>_n_: an Intl mathematical value,</del>
          <ins>_intlMV_: an Intl MV Record,</ins>
        ): a Record with fields [[PluralCategory]] (*"zero"*, *"one"*, *"two"*, *"few"*, *"many"*, or *"other"*) and [[FormattedString]] (a String)
      </h1>
      …
      <emu-alg>
        1. <ins>Let _n_ be _intlMV_.[[Value]].</ins>
```

gibson042:
Fixes #720

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -419,8 +419,8 @@ markEffects: true
       </tr>
       <tr>
         <td><emu-xref href="#variable-declarations" title></emu-xref></td>
-        <td>Inform the linter of the existence of a variable.</td>
-        <td>`[declared="x"]`</td>
+        <td>Inform the linter of the existence or nonexistence of variables.</td>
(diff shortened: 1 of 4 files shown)


gibson042:
#724 resolved the motivating use case for this PR. We can revisit it if another one surfaces.

PR #728, opened 2026-09-01
stricter increment lint

### LINKED ISSUE #725: False positive: Record fields named [[Increment]]
#693 broke Temporal which has a Record representing a requested rounding operation, with [[Precision]], [[Unit]] and [[Increment]] fields. You can guess where this is going:

```
error: prefer "set _x_ to _x_ + 1" over "increment" (algorithm-spelling) at temporal/formatting.emu:218:13:
  216 |           [[Precision]]: ~minute~,
  217 |           [[Unit]]: *"minute"*,
> 218 |           [[Increment]]: 1
      |             ^
  219 |         }.
  220 |       1. If _smallestUnit_ is *"second"*, return the Record {
  221 |           [[Precision]]: 0,
```

bakkot:
Fixes https://github.com/tc39/ecmarkup/issues/725.

### DIFF
--- a/src/lint/rules/algorithm-spelling.ts
+++ b/src/lint/rules/algorithm-spelling.ts
@@ -13,7 +13,7 @@ const matchers = [
     message: 'prefer "set _x_ to _x_ + 1" over "increment"',
   },
   {
-    pattern: /\b[Dd]ecrement\b/gu,
+    pattern: /\b[Dd]ecrement _/gu,
     message: 'prefer "set _x_ to _x_ - 1" over "decrement"',
   },
   {
--- a/src/lint/rules/algorithm-spelling.ts
+++ b/src/lint/rules/algorithm-spelling.ts
@@ -9,7 +9,7 @@ const ruleId = 'algorithm-spelling';
 // Note that these will be composed, so cannot contain backreferences
 const matchers = [
   {
-    pattern: /\b[Ii]ncrement\b/gu,
+    pattern: /\b[Ii]ncrement _/gu,
     message: 'prefer "set _x_ to _x_ + 1" over "increment"',
   },
   {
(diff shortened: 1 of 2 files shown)


src/lint/rules/algorithm-spelling.ts
```diff
@@ -9,7 +9,7 @@ const ruleId = 'algorithm-spelling';
 // Note that these will be composed, so cannot contain backreferences
 const matchers = [
   {
-    pattern: /\b[Ii]ncrement\b/gu,
+    pattern: /\b[Ii]ncrement _/gu,
```
gibson042, line 12: We should apply the same pattern below as well, keeping consistency for "increment"/"decrement"/"increase"/"decrease" checks.
```patch
   {
-    pattern: /\b[Dd]ecrement\b/gu,
+    pattern: /\b[Dd]ecrement _/gu,
     message: 'prefer "set _x_ to _x_ - 1" over "decrement"',
   },
```

bakkot: Disagree: "increment" a noun that's reasonable to use in algorithms e.g. as a record name; "decrement" is not. (I suppose I could imagine "increase" or "decrease" being used as nouns but it'd be a stretch.)

gibson042: [[Decrement]] is absolutely a reasonable record field name, and even if it weren't I'd still be arguing for consistency here to avoid confusion about what sets "decrement" apart—especially since all the cases we care about catching are still covered by `/\b[Dd]ecrement _/gu`.

michaelficarra: I imagine I would object to a record field named `[[Decrement]]`, but I also think we should lint against increment/decrement equally.

bakkot: OK, done.

michaelficarra (approved):

